PR #90, opened 2026-02-02
fix: make gremlin IDs globally unique across files

Kenneth-Javier:
## Summary

Fix duplicate gremlin IDs causing incorrect results in parallel execution mode. Each file was generating gremlin IDs starting from `g001`, causing ID collisions when processing multiple files. The `gremlin_by_id` dict would overwrite entries with duplicate keys, leading to incorrect mutation scores (e.g., 2% killed instead of 73%).

## Changes

- Added `_make_file_prefix()` static method to both `GremlinCollector` and `MutationSwitchingTransformer` classes
- Gremlin IDs now include a file-based prefix: `{filename}_{4-char-hash}_g{num}` (e.g., `utils_a1b2_g001`)
- Updated test to use actual gremlin ID instead of hardcoded `'g001'`
- Added `hashlib` and `pathlib.Path` imports

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## Related Issues

None found - this was discovered through real-world usage.

## Testing

- [x] All existing tests pass
- [ ] Added new tests for this change
- [x] Manually tested

**Manual testing details:**
- Ran parallel mode on a real project (PostService with 5 Python files)
- Before fix: 2% killed (incorrect due to ID collisions)
- After fix: 73% killed (correct, matches sequential mode)
- All 504 small tests pass
- All 30 medium tests pass

## Checklist

- [ ] Tests written FIRST (TDD)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,7 +171,7 @@ jobs:
         run: uv sync --extra dev
 
       - name: Run pip-audit
-        run: uv run pip-audit --ignore-vuln CVE-2026-25990  # pillow <12; blocked by mkdocs-material[imaging] <12.0 cap
+        run: uv run pip-audit
 
   all-checks:
     name: All Checks Passed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,14 +38,12 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "cairosvg>=2.7.0",
     "commitizen>=3.13.0",
     "coverage[toml]>=7.0.0",
     "mkdocs>=1.5.0",
-    "mkdocs-material[imaging]>=9.0.0",
+    "mkdocs-material>=9.0.0",
     "mkdocs-minify-plugin>=0.7.0",
     "mkdocstrings[python]>=0.24.0",
-    "pillow>=10.0.0",
     "mypy>=1.7.0",
     "pip>=26.0",
     "pip-audit>=2.6.0",
@@ -59,12 +57,10 @@ dev = [
     "tox-uv>=1.0.0",
 ]
 docs = [
-    "cairosvg>=2.7.0",
     "mkdocs>=1.5.0",
-    "mkdocs-material[imaging]>=9.0.0",
+    "mkdocs-material>=9.0.0",
     "mkdocs-minify-plugin>=0.7.0",
     "mkdocstrings[python]>=0.24.0",
-    "pillow>=10.0.0",
 ]
 
 [project.urls]
--- a/uv.lock
+++ b/uv.lock
@@ -101,110 +101,12 @@ wheels = [
 ]
 
 [[package]]
-name = "cairocffi"
-version = "1.7.1"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "cffi" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/70/c5/1a4dc131459e68a173cbdab5fad6b524f53f9c1ef7861b7698e998b837cc/cairocffi-1.7.1.tar.gz", hash = "sha256:2e48ee864884ec4a3a34bfa8c9ab9999f688286eb714a15a43ec9d068c36557b", size = 88096, upload-time = "2024-06-18T10:56:06.741Z" }
-wheels = [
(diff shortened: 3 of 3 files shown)


mikelane (approved):
Thank you so much for this contribution, @Kenneth-Javier! Welcome to pytest-gremlins! 🎉

This is a genuinely important bug find. The fact that you discovered this through real-world usage on a multi-file project and traced it all the way down to `gremlin_by_id` dict key collisions is impressive diagnostic work. Going from **2% to 73% kill rate** demonstrates this fix has real, measurable impact.

## Quality Checks

| Check | Result |
|-------|--------|
| `ruff check` | ✅ Passed |
| `ruff format --check` | ✅ Passed |
| `mypy` | ✅ Passed (0 issues, 39 files) |
| `pytest tests/small` | ✅ 438 passed, 1 skipped |

Everything is clean!

## Code Review

**The core fix is correct and well-targeted.** The gremlin ID uniqueness approach (`{stem}_{4-char-hash}_g{counter}`) is sound. I verified that the ID is used as an opaque string throughout the codebase (dict keys, env vars, cache keys, JSON reports) - no code parses the format, so the change is fully transparent to all consumers.

### A few small suggestions (none of these should block merging):

1. **DRY opportunity**: The `_make_file_prefix()` method is identical in both `GremlinCollector` and `MutationSwitchingTransformer`. One thought would be to extract it to a module-level helper function - the module already has several shared functions like `create_gremlins_for_compare`, so the pattern fits. But totally fine to address in a follow-up if you prefer.

2. **Docstring update**: The `Gremlin` dataclass docstring at `gremlin.py:22` still references the old `g001` format. Similarly `build_switching_expression`'s docstring examples. Easy to update whenever convenient.

3. **Test coverage idea**: A dedicated test that processes two different files and asserts their gremlin IDs are distinct would serve as a great regression test for the exact bug this fixes. Happy to help create this as a follow-up issue if you're interested.

### Before merging

The branch is currently behind `main` (we just merged PR #92 which touched `plugin.py`). A rebase will be needed, but since your changes are in `transformer.py` only, conflicts are very unlikely. **Happy to handle the rebase on our side if you'd prefer** - just let us know!

We also need CI checks to pass on the rebased commit before we can merge.

Thanks again for this excellent contribution! 🙌

mikelane:
Hey @Kenneth-Javier! A couple of notes on CI:

### Coverage

The `codecov/patch` check is flagging that the new code doesn't meet our 80% line coverage threshold. Since `_make_file_prefix()` is added to both `GremlinCollector` and `MutationSwitchingTransformer`, we need tests that exercise those paths.

Here are some ideas for tests that would boost coverage and also serve as great regression tests for the exact bug you fixed:

1. **Two files produce distinct gremlin IDs** - the core regression test:
   ```python
   def test_gremlins_from_different_files_have_unique_ids(self):
       source = "x = 1 + 2"
       gremlins_a, _ = transform_source(source, "file_a.py", [...])
       gremlins_b, _ = transform_source(source, "file_b.py", [...])
       ids_a = {g.gremlin_id for g in gremlins_a}
       ids_b = {g.gremlin_id for g in gremlins_b}
       assert ids_a.isdisjoint(ids_b)
   ```

2. **Same-named files in different directories produce distinct IDs**:
   ```python
   def test_same_filename_different_dirs_have_unique_ids(self):
       source = "x = 1 + 2"
       gremlins_a, _ = transform_source(source, "src/utils.py", [...])
       gremlins_b, _ = transform_source(source, "tests/utils.py", [...])
       ids_a = {g.gremlin_id for g in gremlins_a}
       ids_b = {g.gremlin_id for g in gremlins_b}
       assert ids_a.isdisjoint(ids_b)
   ```

3. **ID format includes file prefix** (sanity check):
   ```python
   def test_gremlin_id_includes_file_stem(self):
       source = "x = 1 + 2"
       gremlins, _ = transform_source(source, "my_module.py", [...])
       assert all("my_module" in g.gremlin_id for g in gremlins)
   ```

These would go in `tests/small/instrumentation/test_transformer.py` and would give you full coverage on the new code while also documenting the bug fix for future readers.

### Security Audit

The `Security Audit` failure is a pre-existing issue on our side (pip version pinning) - nothing to do with your changes. We've already fixed it on `main`, so once you rebase, that should clear up.

### Rebase

Speaking of which - `main` has moved forward since you branched (we merged #92). You'll need a rebase before we can merge. If you'd prefer, we're happy to handle the rebase on our end - just let us know!

Thanks again for this great contribution! 🙌

PR #106, opened 2026-02-07
fix: don't count pytest collection/import errors as zapped

### LINKED ISSUE #93: [Bug]: Import errors counted as zapped gremlins (false positives)
### Bug Description

In `_test_gremlin`, any non-zero subprocess exit code is classified as ZAPPED:

```python
if result.returncode != 0:
    return GremlinResult(
        gremlin=gremlin,
        status=GremlinResultStatus.ZAPPED,
        killing_test='unknown',
    )
```

This conflates three different scenarios:

1. **Test actually caught the mutation** → correct ZAPPED
2. **Import error caused by the mutation** (e.g., `SyntaxError`, `NameError`) → **false ZAPPED** (no test ran)
3. **Collection error** (pytest couldn't collect tests) → **false ZAPPED**

### Impact

Mutations that break module imports at load time are marked as "zapped" even though no test actually detected them. This inflates mutation scores and gives false confidence in test suite quality.

This is especially relevant after #91 / #92, where gremlins with zero coverage data now fall back to running all tests. Module-level mutations that cause import errors will crash the subprocess before any test runs, but still get counted as zapped.

### Example

```python
# src/config.py
TIMEOUT = compute_default() - 2  # Gremlin: - to +
```

If `compute_default()` raises during import when the arithmetic mutation is active:
- Subprocess exits with non-zero code
- Plugin classifies it as ZAPPED
- No test actually ran or caught anything

### Expected Behavior

The plugin should distinguish between:
- **Test failure** (exit code from pytest indicating test assertions failed) → ZAPPED
- **Import/collection error** (exit code from Python/pytest indicating startup failure) → a new status like `ERROR` or `IMPORT_ERROR`, not ZAPPED

### Possible Approaches

1. **Parse subprocess stderr/stdout** for import error signatures before classifying
2. **Check pytest exit codes**: pytest uses specific exit codes (e.g., exit code 2 = collection error, exit code 1 = test failure). Only count exit code 1 as ZAPPED
3. **Use a sentinel**: have the bootstrap script write a marker file when tests actually start running; if the marker is absent, the gremlin errored rather than being zapped

### Context

Discovered during investigation of #91. This is a pre-existing issue that affects all operators, not just module-level code.

### Affected Code

- `src/pytest_gremlins/plugin.py` — `_test_gremlin()` function (line ~1412)
- `src/pytest_gremlins/parallel/pool.py` — similar return code logic in worker execution

lunareed720:
Fixes #93.

### What changed
- Treat only exit code 1 (tests failed) as a gremlin being **ZAPPED**; other non-zero exit codes are now **ERROR** (e.g. pytest collection/import/usage/internal errors).
- Normalize collected node IDs for subprocess execution (strip plugin-added suffixes like `[SMALL]` and make absolute paths relative to `rootdir`) so gremlin subprocess runs don't fail with usage/collection errors.

### Tests
- Added unit tests covering exit-code classification in both the single-process runner and the worker pool.

### DIFF
--- a/tests/small/test_plugin_exit_code_classification.py
+++ b/tests/small/test_plugin_exit_code_classification.py
@@ -6,14 +6,18 @@ Issue #93: import/collection errors were being counted as "zapped".
 from __future__ import annotations
 
 import ast
-from pathlib import Path
 import subprocess
+from typing import TYPE_CHECKING
 
 import pytest
 
 from pytest_gremlins.instrumentation.gremlin import Gremlin
 from pytest_gremlins.plugin import _test_gremlin
 from pytest_gremlins.reporting.results import GremlinResultStatus
+
+
+if TYPE_CHECKING:
+    from pathlib import Path
 
 
 @pytest.fixture
@@ -34,7 +38,23 @@ def sample_gremlin() -> Gremlin:
 class TestTestGremlinExitCodeClassification:
     """Tests for _test_gremlin exit code handling."""
 
-    def test_exit_code_1_is_zapped(self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin) -> None:
+    def test_exit_code_0_is_survived(
+        self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin
+    ) -> None:
+        """Exit code 0 means all tests passed (mutation not caught) -> SURVIVED."""
+
+        def fake_run(*_args: object, **_kwargs: object) -> subprocess.CompletedProcess[bytes]:
+            return subprocess.CompletedProcess(args=['pytest'], returncode=0, stdout=b'', stderr=b'')
+
+        monkeypatch.setattr('pytest_gremlins.plugin.subprocess.run', fake_run)
+
+        result = _test_gremlin(sample_gremlin, ['pytest'], tmp_path, instrumented_dir=None)
+
+        assert result.status == GremlinResultStatus.SURVIVED
+
+    def test_exit_code_1_is_zapped(
+        self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin
+    ) -> None:
         """Exit code 1 means tests failed (mutation caught) -> ZAPPED."""
 
         def fake_run(*_args: object, **_kwargs: object) -> subprocess.CompletedProcess[bytes]:
@@ -46,11 +66,26 @@ class TestTestGremlinExitCodeClassification:
 
         assert result.status == GremlinResultStatus.ZAPPED
 
-    def test_exit_code_2_is_error(self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin) -> None:
-        """Exit code 2 indicates pytest errors (collection/import/etc) -> ERROR."""
+    @pytest.mark.parametrize(
+        'exit_code',
+        [
+            pytest.param(2, id='interrupted'),
+            pytest.param(3, id='internal-error'),
+            pytest.param(4, id='usage-error'),
+            pytest.param(5, id='no-tests-collected'),
+        ],
+    )
+    def test_non_test_exit_codes_are_error(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        tmp_path: Path,
+        sample_gremlin: Gremlin,
+        exit_code: int,
+    ) -> None:
+        """Exit codes 2-5 indicate non-test failures -> ERROR."""
 
         def fake_run(*_args: object, **_kwargs: object) -> subprocess.CompletedProcess[bytes]:
-            return subprocess.CompletedProcess(args=['pytest'], returncode=2, stdout=b'', stderr=b'')
+            return subprocess.CompletedProcess(args=['pytest'], returncode=exit_code, stdout=b'', stderr=b'')
 
         monkeypatch.setattr('pytest_gremlins.plugin.subprocess.run', fake_run)
 
--- a/src/pytest_gremlins/plugin.py
+++ b/src/pytest_gremlins/plugin.py
@@ -286,7 +286,7 @@ def pytest_collection_finish(session: pytest.Session) -> None:
     node_ids = [item.nodeid for item in session.items]
     normalized_node_ids = _make_node_ids_relative(node_ids, rootdir)
     gremlin_session.test_node_ids = {
-        item.name: node_id for item, node_id in zip(session.items, normalized_node_ids)
+        item.name: node_id for item, node_id in zip(session.items, normalized_node_ids, strict=True)
     }
 
     source_files = _discover_source_files(session, gremlin_session)


tests/small/test_plugin_exit_code_classification.py
```diff
@@ -0,0 +1,59 @@
...
+        operator_name='ComparisonOperatorSwap',
+        description='> to >=',
+    )
+
+
+@pytest.mark.small
+class TestTestGremlinExitCodeClassification:
+    """Tests for _test_gremlin exit code handling."""
+
+    def test_exit_code_1_is_zapped(self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin) -> None:
```
mikelane: **[E501] Line too long (123 > 120)** -- Both method signatures in this class exceed the project's 120-character line limit. `ruff format` will wrap these automatically. The fix looks like:

```python
def test_exit_code_1_is_zapped(
    self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin
) -> None:
```

You can fix both at once with: `uv run ruff format tests/small/test_plugin_exit_code_classification.py`

tests/small/test_plugin_exit_code_classification.py
```diff
@@ -0,0 +1,59 @@
...
+        """Exit code 2 indicates pytest errors (collection/import/etc) -> ERROR."""
+
+        def fake_run(*_args: object, **_kwargs: object) -> subprocess.CompletedProcess[bytes]:
+            return subprocess.CompletedProcess(args=['pytest'], returncode=2, stdout=b'', stderr=b'')
+
+        monkeypatch.setattr('pytest_gremlins.plugin.subprocess.run', fake_run)
+
+        result = _test_gremlin(sample_gremlin, ['pytest'], tmp_path, instrumented_dir=None)
+
+        assert result.status == GremlinResultStatus.ERROR
```
mikelane: Nice work testing _both_ directions of the exit-code classification (1 = ZAPPED, 2 = ERROR). The monkeypatch approach is clean and focused.

**Suggestion: Consider adding a test for exit code 0 (SURVIVED).**  This would complete the three-way branching logic and make it easy for future maintainers to see all three cases in one place. Something like:

```python
def test_exit_code_0_is_survived(
    self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, sample_gremlin: Gremlin
) -> None:
    """Exit code 0 means tests passed (mutation survived) -> SURVIVED."""

    def fake_run(*_args: object, **_kwargs: object) -> subprocess.CompletedProcess[bytes]:
        return subprocess.CompletedProcess(args=['pytest'], returncode=0, stdout=b'', stderr=b'')

    monkeypatch.setattr('pytest_gremlins.plugin.subprocess.run', fake_run)

    result = _test_gremlin(sample_gremlin, ['pytest'], tmp_path, instrumented_dir=None)

    assert result.status == GremlinResultStatus.SURVIVED
```

You could also consider `pytest.mark.parametrize` to cover multiple non-test exit codes (3, 4, 5 for pytest's other exit codes) in a single test. Not required -- just a thought for thoroughness!

**Also:** You might want to add a test for `exit code 5` (no tests collected) since that is a common real-world scenario when coverage-guided selection selects zero tests. Currently that would be classified as ERROR, which seems correct, but having a test documenting that decision would be valuable.

mikelane (changes requested):
Welcome to the project, @lunareed720! Thank you for tackling this bug -- issue #93 is a real correctness problem and your fix addresses it head-on. This is a well-structured first contribution.

---

## Review Summary

### What you did well

- **Correct diagnosis of the root cause.** The old `if result.returncode != 0` logic was genuinely wrong -- it conflated pytest test failures (exit 1) with collection/import/internal errors (exit 2, 3, 4, 5). Your fix correctly dispatches on `== 0`, `== 1`, and "everything else", which aligns with [pytest's documented exit codes](https://docs.pytest.org/en/stable/reference/exit-codes.html).

- **Both code paths fixed.** You updated both `_test_gremlin()` in `plugin.py` (single-process runner) AND `_run_gremlin_test()` in `parallel/pool.py` (worker pool). The issue called out both locations and you caught them both. That attention to detail is appreciated.

- **Meaningful, well-named tests.** Your test names are declarative (`test_exit_code_1_is_zapped`, `test_exit_code_2_is_error`, `test_non_test_exit_code_returns_error_status`) -- no "should" phrasing. The docstrings add useful context. The monkeypatch approach in the plugin test is clean and the `sys.exit(2)` approach in the pool test is pragmatic.

- **Node ID normalization.** The `_make_node_ids_relative` call in `pytest_collection_finish` is a sensible addition that addresses a real problem with absolute paths and plugin-added suffixes breaking subprocess execution.

- **The existing `GremlinResultStatus.ERROR` enum value was already in place** and you correctly used it rather than inventing a new status. Good use of the existing domain model.

### Blockers (must fix before merge)

**1. Ruff lint failures (4 violations)**

The CI will fail on these. Running `uv run ruff check src tests` produces 4 errors:

| File | Rule | Issue |
|------|------|-------|
| `src/pytest_gremlins/plugin.py:286` | B905 | `zip()` needs `strict=True` |
| `tests/small/test_plugin_exit_code_classification.py:9` | TC003 | `Path` import should be in `TYPE_CHECKING` block |
| `tests/small/test_plugin_exit_code_classification.py:37` | E501 | Line too long (123 > 120) |
| `tests/small/test_plugin_exit_code_classification.py:49` | E501 | Line too long (122 > 120) |

**Quick fix:** Run `uv run ruff format src tests` to auto-fix the line-length and formatting issues, then manually add `strict=True` to the `zip()` call and move `Path` into a `TYPE_CHECKING` block. See inline comments for exact fixes.

**2. Formatting violations (2 files)**

`uv run ruff format --check src tests` reports that `plugin.py` and `test_plugin_exit_code_classification.py` need reformatting. Running `uv run ruff format src tests` will fix these automatically.

### Scope concern (not a blocker, but worth discussing)

**`pyproject.toml` coverage config changes are bundled in.** This PR adds significant coverage configuration changes: new `omit` patterns, new `exclude_lines`, new `exclude_also`, changes `fail_under` from 60 to 69, and changes `skip_covered`. These changes seem orthogonal to the exit-code classification fix.

I'd recommend splitting these into a separate PR to keep this one focused on the bug fix. It makes the review cleaner and the git history easier to reason about. If the coverage changes are needed to make CI pass with this fix, a note explaining that dependency would be helpful.

### Suggestions (non-blocking)

- **Add a test for exit code 0 (SURVIVED)** in `test_plugin_exit_code_classification.py` to complete the three-way branch coverage. See inline comment for example code.
- **Consider `pytest.mark.parametrize`** for testing multiple non-test exit codes (2, 3, 4, 5) -- this would document pytest's full exit code spectrum in one readable test.
- **Missing `@pytest.mark.small`** on `TestTestGremlinExitCodeClassification` -- while the `tests/small/` directory might handle this automatically via conftest, the class already has it (good!), so this is fine. Just noting for consistency.

### Quality gauntlet results

| Check | Result |
|-------|--------|
| `uv run ruff check src tests` | FAIL (4 violations) |
| `uv run ruff format --check src tests` | FAIL (2 files) |
| `uv run mypy src/pytest_gremlins` | PASS (0 issues in 39 files) |
| `uv run pytest tests/small -m small` | PASS (449 passed, 1 skipped) |
| Config integrity | pyproject.toml changes bundled (see above) |

### Verdict: CHANGES REQUESTED

The core logic is correct and well-tested. The fix properly addresses issue #93. Please fix the 4 ruff lint violations and 2 formatting issues, and we should be good to go. The coverage config changes are a conversation to have with the maintainer (@mikelane) about whether to keep them bundled or split out.

Again, great first contribution -- looking forward to seeing the updated PR!

PR #175, opened 2026-02-20
feat: integrate pytest-xdist -n flag for parallel mutation execution (#130)

### LINKED ISSUE #130: feat: integrate with pytest-xdist -n flag for parallelism control
## Summary

pytest-gremlins currently requires `--gremlin-parallel --gremlin-workers N` to enable
parallel mutation testing. This is verbose and duplicates functionality that pytest-xdist
already provides via `-n`.

## Proposed Change

Hook into pytest-xdist's `-n` flag as the canonical way to control parallelism in
pytest-gremlins:

- If pytest-xdist is installed and `-n auto` or `-n N` is passed, pytest-gremlins uses
  that worker count for parallel mutation execution
- `-n` becomes the single flag to enable and configure parallelism
- `--gremlin-parallel` and `--gremlin-workers` are deprecated (kept for backwards
  compatibility but no longer needed)

## User-Facing Change

**Before:**
```bash
pytest --gremlins --gremlin-parallel --gremlin-workers 8
```

**After:**
```bash
pytest --gremlins -n auto
pytest --gremlins -n 4
```

## Implementation Notes

- Check `config.option.numprocesses` (xdist's internal name for `-n`) in `pytest_configure`
- If xdist is not installed, `-n` is unavailable — existing `--gremlin-parallel` fallback
  remains
- `auto` maps to `os.cpu_count()`
- Parallelism should only activate for the mutation phase, not the initial test collection
  (xdist would interfere with collection if actually enabled for the whole run — need to
  intercept before xdist activates, or detect the flag without letting xdist run)

## Open Question

Does letting xdist actually run conflict with gremlins' own subprocess-based parallelism?
If so, we may need to detect `-n` early and disable xdist from running while still using
the value for our own worker pool.

mikelane:
## Summary

- Reads xdist's `-n` flag (`config.option.numprocesses`) and uses it as the worker count for pytest-gremlins' mutation subprocess pool
- `-n auto` maps to all CPU cores; `-n 4` uses 4 workers; omitting `-n` keeps sequential mode
- `--gremlin-parallel` and `--gremlin-workers` are deprecated — they still work but emit `PytestDeprecationWarning` when xdist is also present
- Docs updated across configuration guide, cookbook, and architecture pages to reflect the new interface

Closes #130

## Test plan

- [ ] `uv run pytest tests/small -m small` — unit tests for `_resolve_parallel_from_xdist` and `_read_parallel_config`
- [ ] `uv run pytest tests/medium` — integration tests for config wiring and deprecation warnings
- [ ] `pytest --gremlins -n auto` in a project with xdist installed — parallel workers start
- [ ] `pytest --gremlins -n 1` — sequential fallback, no workers
- [ ] `pytest --gremlins --gremlin-parallel --gremlin-workers=4` with xdist present — deprecation warning emitted
- [ ] `pytest --gremlins --gremlin-parallel` without xdist — still works, no warning

### DIFF
--- a/tests/small/test_plugin_helpers.py
+++ b/tests/small/test_plugin_helpers.py
@@ -297,6 +297,13 @@ class TestReadParallelConfig:
         assert enabled is True
         assert workers is None
 
+    def test_xdist_installed_but_n_not_passed_falls_back_to_gremlin_parallel(self) -> None:
+        """xdist installed but -n not passed: numprocesses=None, old flags apply."""
+        config = self._make_config(gremlin_parallel=True, numprocesses=None)
+        enabled, workers = _read_parallel_config(config)
+        assert enabled is True
+        assert workers is None
+
 
 @pytest.mark.small
 class TestBuildTestCommand:


tests/small/test_plugin_helpers.py
```diff
@@ -182,6 +189,115 @@ def test_file_not_relative_to_rootdir(self, tmp_path: Path) -> None:
...
+        assert enabled is False
+        assert workers is None
+
+    def test_xdist_not_installed_gremlin_parallel_enables_parallel(self) -> None:
+        """Without xdist, --gremlin-parallel still works."""
+        config = self._make_config(gremlin_parallel=True)
+        enabled, workers = _read_parallel_config(config)
+        assert enabled is True
+        assert workers is None
+
```
mikelane: The five new tests cover B1 (xdist absent), B2 (warning fires), B3 (xdist present, no old flags), and B4 (numprocesses is not None). Branch B5 — xdist is installed but the user did not pass `-n`, so `numprocesses` is `None` — has no test.

In this scenario the fallback to `gremlin_parallel` / `gremlin_workers` is the correct and intended behaviour, but there is nothing that would catch a regression there. The scenario is reachable in production any time a project has xdist installed but simply does not pass `-n` in a given run.

A test like the one below would close the gap:

```python
def test_xdist_installed_but_n_not_passed_falls_back_to_gremlin_parallel(self) -> None:
    """xdist present with numprocesses=None falls back to --gremlin-parallel."""
    config = self._make_config(gremlin_parallel=True, numprocesses=None)
    enabled, workers = _read_parallel_config(config)
    assert enabled is True
    assert workers is None
```

This is a triangulation gap, not a correctness defect — the production code is right. Worth adding before merge so the branch is pinned.

PR #186, opened 2026-02-21
fix(ci): enable subprocess coverage tracking via sitecustomize.py

mikelane:
## Summary

- The release workflow was failing with 83% coverage against the 97% threshold
- Root cause: large tests spawn subprocess pytest instances via pytester; those subprocess-executed lines in `plugin.py` are invisible to the parent `coverage.py` process
- Fix: add `sitecustomize.py` at the project root calling `coverage.process_startup()`, and set `COVERAGE_PROCESS_START=pyproject.toml` + `PYTHONPATH=.` in both CI and release workflows
- Every subprocess Python instance now writes its own `.coverage.<pid>` file; `coverage combine` merges them before reporting

## Files changed

- **`sitecustomize.py`** (new) — calls `coverage.process_startup()` when `COVERAGE_PROCESS_START` is set
- **`.github/workflows/ci.yml`** — adds `COVERAGE_PROCESS_START` and `PYTHONPATH` env vars to the test step
- **`.github/workflows/release.yml`** — same, for the full test run step
- **`tests/small/test_subprocess_coverage.py`** (new) — verifies `sitecustomize.py` exists, is valid Python, calls `coverage.process_startup()`, and that subprocesses actually write `.coverage.*` files when env vars are set

## Test plan

- [ ] All 7 new tests pass locally
- [ ] ruff, mypy, pre-push hooks all green
- [ ] CI passes on this PR (coverage should be ≥ 97% with subprocess tracking active)
- [ ] Release workflow passes for the next tag

### DIFF
--- a/tests/small/test_subprocess_coverage.py
+++ b/tests/small/test_subprocess_coverage.py
@@ -41,13 +41,15 @@ class TestSitecustomizeExists:
         source = (project_root / 'sitecustomize.py').read_text(encoding='utf-8')
         tree = ast.parse(source)
 
-        # Find any call to coverage.process_startup()
+        # Find calls specifically to coverage.process_startup() (not just anything.process_startup())
         calls = [
             node
             for node in ast.walk(tree)
             if isinstance(node, ast.Call)
             and isinstance(node.func, ast.Attribute)
             and node.func.attr == 'process_startup'
+            and isinstance(node.func.value, ast.Name)
+            and node.func.value.id == 'coverage'
         ]
         assert calls, 'sitecustomize.py must call coverage.process_startup()'
 
@@ -87,14 +89,16 @@ class TestCoverageRunConfig:
         )
 
 
-@pytest.mark.small
+@pytest.mark.medium
 class TestSubprocessCoverageIntegration:
-    """Verify subprocess Python instances write coverage data when env vars are set."""
+    """Verify subprocess Python instances write coverage data when env vars are set.
+
+    Marked medium: these tests spawn real subprocesses with filesystem I/O.
+    """
 
     def test_subprocess_writes_coverage_file_when_env_vars_set(self, tmp_path: Path) -> None:
         project_root = Path(__file__).parent.parent.parent
 
-        # Write a trivial Python module that subprocess will import and run
         target = tmp_path / 'target.py'
         target.write_text('def add(a, b):\n    return a + b\n\nresult = add(1, 2)\n')
 
@@ -117,3 +121,24 @@ class TestSubprocessCoverageIntegration:
         assert coverage_files, (
             'Subprocess must write a .coverage.* file when COVERAGE_PROCESS_START and PYTHONPATH are set'
         )
+
+    def test_subprocess_does_not_write_coverage_file_without_env_vars(self, tmp_path: Path) -> None:
+        target = tmp_path / 'target.py'
+        target.write_text('def add(a, b):\n    return a + b\n\nresult = add(1, 2)\n')
+
+        env = os.environ.copy()
+        env.pop('COVERAGE_PROCESS_START', None)
+
+        result = subprocess.run(
+            [sys.executable, str(target)],
+            capture_output=True,
+            text=True,
+            cwd=str(tmp_path),
+            env=env,
+            check=False,
+        )
+
+        assert result.returncode == 0, f'Subprocess failed: {result.stderr}'
+
+        coverage_files = list(tmp_path.glob('.coverage*'))
+        assert not coverage_files, 'Subprocess must NOT write .coverage.* files when COVERAGE_PROCESS_START is absent'
--- a/src/pytest_gremlins/parallel/pool.py
+++ b/src/pytest_gremlins/parallel/pool.py
@@ -232,6 +232,10 @@ class WorkerPool:
         if instrumented_dir is not None:
             all_env_vars['PYTEST_GREMLINS_SOURCES_FILE'] = str(Path(instrumented_dir) / 'sources.json')
 
+        # Suppress subprocess coverage tracking: sitecustomize.py fires at Python startup
+        # before --no-cov takes effect, so we explicitly clear the env var here.
+        all_env_vars['COVERAGE_PROCESS_START'] = ''
+
         return self._executor.submit(
             _run_gremlin_test,
             gremlin_id,
--- a/tests/small/parallel/test_pool.py
+++ b/tests/small/parallel/test_pool.py
@@ -230,6 +230,37 @@ class TestWorkerPoolExecution:
         finally:
             Path(script_path).unlink()
 
+    def test_coverage_process_start_is_suppressed_in_subprocess(
+        self,
+        tmp_path: Path,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        """COVERAGE_PROCESS_START is not inherited by mutation subprocesses.
+
+        sitecustomize.py fires at Python startup before --no-cov takes effect,
+        so it must be explicitly suppressed in the subprocess environment.
+        """
+        monkeypatch.setenv('COVERAGE_PROCESS_START', '/some/pyproject.toml')
+
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.py', delete=False) as f:
+            f.write('import os, sys\ncps = os.environ.get("COVERAGE_PROCESS_START", "")\nsys.exit(1 if cps else 0)\n')
+            script_path = f.name
+
+        try:
+            with WorkerPool(max_workers=1, timeout=5) as pool:
+                future = pool.submit(
+                    gremlin_id='g001',
+                    test_command=[sys.executable, script_path],
+                    rootdir=str(tmp_path),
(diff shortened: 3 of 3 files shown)


tests/small/test_subprocess_coverage.py
```diff
@@ -0,0 +1,119 @@
...
+            env=env,
+            check=False,
+        )
+
+        assert result.returncode == 0, f'Subprocess failed: {result.stderr}'
+
+        coverage_files = list(tmp_path.glob('.coverage*'))
+        assert coverage_files, (
+            'Subprocess must write a .coverage.* file when COVERAGE_PROCESS_START and PYTHONPATH are set'
+        )
```
mikelane: This test confirms the positive case well: when both env vars are set, a `.coverage.*` file appears. What is missing is the symmetric negative case — that *without* the env vars, no `.coverage.*` file is written.

That negative test is important for triangulation. Without it, a bug in the mechanism that caused it to *always* write coverage files (regardless of env vars) would be invisible to the test suite. The positive test would still pass, but the semantic contract — 'env vars control whether coverage is collected' — would be violated silently.

I confirmed locally that the negative behavior is correct (no file is written when the vars are absent), so this is not a bug, but the contract is untested. A second test case alongside this one would complete the triangulation.

.github/workflows/ci.yml
```diff
@@ -112,6 +112,9 @@ jobs:
         run: uv sync
 
       - name: Run tests with coverage
+        env:
+          COVERAGE_PROCESS_START: pyproject.toml
+          PYTHONPATH: .
```
mikelane: Setting `COVERAGE_PROCESS_START` and `PYTHONPATH=.` at the job level means every subprocess spawned during the test run inherits them — including the mutation subprocesses that pytest-gremlins spawns to test gremlins.

Looking at `plugin.py`, `_test_gremlin` builds its env with `os.environ.copy()` and then adds `ACTIVE_GREMLIN` and `GREMLIN_SOURCES_FILE`. It already passes `--no-cov` to suppress pytest-cov inside mutation subprocesses, but `COVERAGE_PROCESS_START` is a coverage.py mechanism that operates independently of pytest-cov — it fires through `sitecustomize.py` at Python startup, before pytest even loads. So mutation subprocesses will silently write `.coverage.<pid>` files alongside any other output they produce.

The practical consequences:
1. The `coverage combine` step will attempt to include mutation subprocess data, polluting the coverage report with lines executed under arbitrary mutations rather than the clean source.
2. Each mutation subprocess starts slightly slower (coverage.py initialises, opens a data file, registers an atexit handler). Given that this project's performance rationale is explicitly speed-first, this is worth noting.
3. The existing suppression logic (`--no-cov`, `addopts=`) does not clear `COVERAGE_PROCESS_START` from the subprocess environment.

The existing `ba6a3cc` commit ('suppress addopts and coverage in mutation subprocesses') suggests this interaction has been thought about before. It may be worth extending that suppression to also unset `COVERAGE_PROCESS_START` in the env dict built inside `_test_gremlin` (and the parallel pool equivalent in `pool.py`). That would make the mutation execution path immune to however the CI environment is configured.

sitecustomize.py
```diff
@@ -0,0 +1,14 @@
...
+writes its own .coverage.<pid> data file. The parent process combines
+them with ``coverage combine``.
+
+Reference: https://coverage.readthedocs.io/en/latest/subprocess.html
+"""
+
+import coverage
+
+
+coverage.process_startup()
```
mikelane: This file is not excluded from the package distribution. `pyproject.toml` configures the wheel to include only `src/pytest_gremlins` (via `[tool.hatch.build.targets.wheel] packages = ["src/pytest_gremlins"]`), so the wheel is clean. However, the sdist `include` list in `[tool.hatch.build.targets.sdist]` does not explicitly enumerate root-level files — Hatchling's sdist includes everything not gitignored by default, so `sitecustomize.py` will be present in the sdist.

This is worth a conscious decision: if a downstream user installs from the sdist and happens to have `." on their `PYTHONPATH`, Python will import this project's `sitecustomize.py` into their environment. `sitecustomize.py` is a well-known Python mechanism (exactly one can be active per interpreter startup), so shipping it in an sdist can silently override whatever `sitecustomize.py` the user already has if import path ordering places this one first.

The standard mitigation is to add `sitecustomize.py` to `.gitignore` (keeping it local-only, never committed) and generate it as part of the CI setup step, or to add it to the sdist exclusion list. Neither is the right answer for every project — but the current state silently ships it to anyone who installs the sdist.

mikelane (commented):
## Review: fix(ci): enable subprocess coverage tracking via sitecustomize.py

### What this PR does well

The core mechanism is correct and idiomatic. `coverage.process_startup()` is the documented approach for subprocess coverage tracking, and the combination of `COVERAGE_PROCESS_START=pyproject.toml` plus `PYTHONPATH=.` is exactly how the coverage.py docs recommend wiring it. The behavioral integration test (`test_subprocess_writes_coverage_file_when_env_vars_set`) is the right shape — it spawns a real subprocess and checks for a real file on disk. I confirmed locally that the negative case (no env vars → no file written) is also correct at runtime.

---

### Issues found

**1. MEDIUM — Mutation subprocesses inherit `COVERAGE_PROCESS_START` (correctness + performance)**

Setting these env vars at the job level means every subprocess that inherits the environment will activate coverage tracking — including the mutation subprocesses pytest-gremlins spawns per gremlin. `_test_gremlin` in `plugin.py` builds its env with `os.environ.copy()` and does not clear `COVERAGE_PROCESS_START`. The `--no-cov` flag already passed to mutation subprocesses suppresses pytest-cov, but `COVERAGE_PROCESS_START` fires through `sitecustomize.py` at Python startup before pytest loads — `--no-cov` does not stop it.

Consequences: mutation subprocess data contaminates the `coverage combine` results, and every mutation subprocess pays the cost of coverage initialisation (file open, atexit registration). Commit `ba6a3cc` already set a precedent for suppressing coverage inside mutation subprocesses — that suppression should extend to unsetting `COVERAGE_PROCESS_START` in the env dict. Inline comment on `ci.yml:117`.

**2. MEDIUM — `sitecustomize.py` is shipped in the sdist**

`sitecustomize.py` is not excluded from the Hatchling sdist (which includes all non-gitignored root files by default). The wheel is clean. But a user who installs from the sdist and has `.` on their `PYTHONPATH` will get this file imported into their interpreter on startup — potentially silencing their own `sitecustomize.py`. The standard approach is either to gitignore the file and generate it in CI, or to explicitly exclude it from the sdist manifest. Inline comment on `sitecustomize.py:14`.

**3. LOW — AST tests do not verify receiver identity**

`test_sitecustomize_calls_coverage_process_startup` finds any AST node where `.attr == 'process_startup'` — it does not check that `node.func.value.id == 'coverage'`. An implementation that calls `anything.process_startup()` passes the test. The import and call tests are also unlinked, so an implementation that imports `coverage` but calls `other.process_startup()` passes both. The behavioral test already provides the real confidence here; if these static tests are kept, tightening the AST predicate would make them meaningful. Inline comment on `test_subprocess_coverage.py:52`.

**4. LOW — String-search config tests are fragile**

`'parallel = true' in content` and `'source = ["src/pytest_gremlins"]' in content` are literal substring matches. They would pass if the text appears in a comment or a different TOML section, and break if spacing changes (e.g., `source = [ "src/pytest_gremlins" ]`). `tomllib` is in stdlib from Python 3.11 (this project's minimum) and would make these assertions structurally correct. Inline comment on `test_subprocess_coverage.py:77`.

**5. LOW — Missing negative triangulation test**

There is no test that WITHOUT the env vars, no `.coverage.*` file is written. The positive case is covered. The negative case is the other half of the behavioral contract. Without it, a regression where coverage is always written regardless of env vars would be invisible. Inline comment on `test_subprocess_coverage.py:119`.

**6. LOW — Test size classification**

`test_subprocess_writes_coverage_file_when_env_vars_set` is marked `small` but spawns a real subprocess with filesystem I/O. Per the project's Google test size taxonomy (documented in CLAUDE.md), tests with subprocess execution and file I/O are `medium`. It passes in 0.2s today but the classification is wrong and could mislead future contributors about what constraints apply.

---

### No issues found

- `coverage.process_startup()` is a documented no-op when `COVERAGE_PROCESS_START` is unset (confirmed at runtime). No risk of interference when the env var is absent.
- `PYTHONPATH=.` on the GitHub Actions runner (checkout to workspace root) is safe — `.` resolves to the workspace, which is where `sitecustomize.py` lives. No import shadowing risk in that controlled environment.
- Linting (ruff), formatting, and mypy all pass clean on the new files.
- `coverage.process_startup()` has no security concerns; it reads a config file path from an env var you control and starts coverage measurement. There is no code execution, no network access, and no privilege escalation involved.

---

### Summary

The mechanism is sound and the CI wiring is correct. The most substantive issue is issue 1 (mutation subprocesses inherit `COVERAGE_PROCESS_START`) — that has real consequences for report correctness and is in tension with the project's speed-first design principle. Issue 2 (sdist) is a quieter correctness concern worth resolving before a public release. Issues 3–6 are test quality improvements that would make the suite more robust to future changes.

PR #210, opened 2026-02-27
feat(release): adopt frequent stable releases (FastAPI/Ruff model)

### LINKED ISSUE #36: Implement release train workflow (commitizen)
## Context
Manual version management is error-prone: wrong bump type, forgotten changelog entries, tags that don't match `pyproject.toml`. Commitizen automates this entirely — it reads the conventional commit history since the last tag, determines the correct semver bump (patch/minor/major), updates the changelog, and pushes the tag that triggers the publish workflow.

Parent epic: #32
Blocked by: #35

## User Story
As a maintainer, I want version bumps and changelog updates to happen automatically after merges to main, so that I never accidentally ship the wrong version number or miss a changelog entry.

## Acceptance Criteria
- [ ] Workflow triggers on push to `main` (after merge queue approves the PR)
- [ ] A `bump:` commit from this workflow does NOT trigger the workflow again (infinite loop prevention verified)
- [ ] `feat:` commits produce a minor version bump
- [ ] `fix:` commits produce a patch version bump
- [ ] `feat!:` or `BREAKING CHANGE:` in body produces a major bump
- [ ] `docs:`, `chore:`, `refactor:`, `test:`, `ci:` commits produce no release
- [ ] `CHANGELOG.md` is updated with the new version section and entries
- [ ] An annotated tag (not lightweight) is pushed for every release
- [ ] If no releasable commits exist since the last tag, the workflow exits cleanly with no version bump
- [ ] The workflow uses a GitHub App token (not `GITHUB_TOKEN`) so the bump commit does not trigger other workflows
- [ ] Concurrent runs are prevented (only one release train at a time)

## Definition of Done
- [ ] Tests updated/added (where applicable)
- [ ] No regressions in existing tests
- [ ] mypy passes with no new ignores
- [ ] ruff passes with no new violations
- [ ] GitHub App created (`pytest-gremlins-release-bot`) and secrets `RELEASE_APP_ID` / `RELEASE_APP_PRIVATE_KEY` stored
- [ ] Workflow committed to `.github/workflows/release-train.yml`
- [ ] `pyproject.toml` commitizen config verified (annotated tags, semver, incremental changelog)
- [ ] End-to-end verified: merge `feat:` PR → version bumped → tag pushed → publish workflow triggered

## Notes
Do not use `--no-verify` on any git command. Do not amend commits. If a tag already exists for the target version, the workflow must fail loudly — it indicates something has gone wrong.

## Blocked by
#35

mikelane:
## Summary

- Add `scripts/release.sh` — local one-liner: show commits → bump version → push tag → triggers `release.yml`
- Add `.github/workflows/cut-release.yml` — same thing from the GitHub Actions UI (choose `auto`/`patch`/`minor`/`major`)
- Replace the RC-heavy release process in `AGENT_GUIDELINES.md` with a "release often" philosophy
- Harden both release tools: empty-release guard, branch/dirty-tree preflight, trap handler, PAT expiry diagnostics, benchmark regression annotation

## Motivation

The old process (manual `cz bump` → push → wait) was high-friction enough that releases were batched. This makes releasing after every few merged PRs a one-liner. Replaces the planned auto-release train (#36) with a manual but equally low-friction approach.

## How to Release (after this merges)

```bash
# From main, with a clean working tree
./scripts/release.sh           # auto-detect bump type
./scripts/release.sh --patch   # or force patch/minor/major
```

Or trigger from **Actions → Cut Release → Run workflow**.

## Phase 2 prerequisite (GitHub Actions path)

Create a fine-grained PAT with **Repository permissions → Contents: Read and write**, store it as `RELEASE_PAT` repository secret, set a 90-day expiry.

## Changes

- `scripts/release.sh` — new
- `.github/workflows/cut-release.yml` — new
- `.github/workflows/release.yml` — benchmark regression now annotated in Actions UI (was silently swallowed by `continue-on-error: true`)
- `docs/design/AGENT_GUIDELINES.md` — Release Process section rewritten
- `CLAUDE.md` — key files updated

## Test plan

- [ ] Merge this PR
- [ ] Run `./scripts/release.sh --patch` from `main` — confirm it shows commits, bumps version, pushes tag, and `release.yml` triggers
- [ ] Confirm `pip install pytest-gremlins==<new-version>` works after pipeline completes
- [ ] (Phase 2) Add `RELEASE_PAT` secret, trigger **Cut Release** from Actions UI, confirm same result

Closes #36

---

### DIFF
--- a/.github/workflows/cut-release.yml
+++ b/.github/workflows/cut-release.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       bump:
-        description: 'Version bump type'
+        description: 'How to bump the version. "auto" detects bump type from conventional commits (recommended).'
         required: false
         default: 'auto'
         type: choice
@@ -51,8 +51,8 @@ jobs:
           if [[ -n "$last_tag" ]]; then
             commit_count=$(git rev-list --count "${last_tag}..HEAD")
             if [[ "$commit_count" -eq 0 ]]; then
-              echo "::error::No commits since last tag (${last_tag}). Nothing to release."
-              echo "No commits since **${last_tag}**. Nothing to release." >> "$GITHUB_STEP_SUMMARY"
+              echo "::error::No commits since ${last_tag}. Merge new PRs to main before cutting a release."
+              echo "No commits since **${last_tag}**. Merge new PRs to main before cutting a release." >> "$GITHUB_STEP_SUMMARY"
               exit 1
             fi
             echo "${commit_count} commits since ${last_tag}:" >> "$GITHUB_STEP_SUMMARY"
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -69,7 +69,7 @@ if [[ "$branch" != "main" ]]; then
   exit 1
 fi
 if ! git diff --quiet || ! git diff --cached --quiet; then
-  echo "Error: Working tree has uncommitted changes. Commit or stash them first." >&2
+  echo "Error: You have uncommitted changes. Commit or stash them first." >&2
   git status --short >&2
   exit 1
 fi
@@ -84,7 +84,7 @@ last_tag=$(git describe --tags --abbrev=0 2>/dev/null || echo "")
 if [[ -n "$last_tag" ]]; then
   commit_count=$(git rev-list --count "${last_tag}..HEAD")
   if [[ "$commit_count" -eq 0 ]]; then
-    echo "Error: No commits since last tag (${last_tag}). Nothing to release." >&2
+    echo "Error: No commits since ${last_tag}. Merge new changes to main before releasing." >&2
     exit 1
   fi
   echo "  (${commit_count} commits since ${last_tag})"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,8 +39,7 @@ jobs:
           PYTHONPATH: .
         run: |
           uv run coverage run -m pytest tests -m "small or medium" -v
-          # coverage combine may exit non-zero when there is only a single data file
-          # (no subprocess coverage to combine). This is expected -- not an error.
+          # Expected: coverage combine exits non-zero with a single data file.
           uv run coverage combine --quiet || echo "::notice::coverage combine found nothing to combine (single data file). This is normal."
           uv run coverage report --show-missing
 
@@ -99,7 +98,7 @@ jobs:
       - name: Annotate benchmark regression
         if: steps.check_regression.outcome == 'failure'
         run: |
-          echo "::warning::Benchmark regression detected (>10% threshold). Review regression_report.txt in the uploaded artifacts."
+          echo "::warning::Benchmark regression detected (>10% threshold). Review regression_report.txt in the 'benchmark-results' artifact."
           echo "### Benchmark Regression Detected" >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo "The benchmark regression check exceeded the 10% threshold." >> "$GITHUB_STEP_SUMMARY"


scripts/release.sh
```diff
@@ -0,0 +1,29 @@
...
+
+INCREMENT_FLAG=""
+case "${1:-auto}" in
+  --patch) INCREMENT_FLAG="--increment PATCH" ;;
+  --minor) INCREMENT_FLAG="--increment MINOR" ;;
+  --major) INCREMENT_FLAG="--increment MAJOR" ;;
+  auto)    INCREMENT_FLAG="" ;;
+  *)       echo "Usage: $0 [--patch|--minor|--major]"; exit 1 ;;
+esac
+
```
mikelane: The `cleanup` name is misleading here. The function performs no cleanup — no temp files to delete, no lock files to release, no resources to free. What it actually does is report failure context on exit.

A name like `report_on_exit` or `on_exit_handler` would match the implementation. `cleanup` creates a mental gap: readers scanning the trap setup expect teardown logic, but find diagnostic printing instead.

This is a low-friction rename: the function and the `trap cleanup EXIT` line. Worth fixing before this becomes a convention others copy.

.github/workflows/cut-release.yml
```diff
@@ -0,0 +1,77 @@
...
+          token: ${{ secrets.RELEASE_PAT }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
```
mikelane: LOW - `uv sync --all-groups` installs test, dev, and docs dependency groups that are not used by this workflow. The only tool needed here is `commitizen` (via `uv run cz`), which belongs to the dev group at most.

Consider whether `uv sync` (no flags, installs default groups only) or a targeted `uv sync --group dev` would be sufficient. Installing 50+ test/benchmark dependencies into a release runner adds ~30s of unnecessary setup time on every trigger and widens the attack surface for supply-chain issues.

Not a blocker — the functionality is correct — but worth tightening before this workflow runs routinely.

.github/workflows/cut-release.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Bump version (minor)
+        if: inputs.bump == 'minor'
+        run: uv run cz bump --yes --increment MINOR
+
+      - name: Bump version (major)
+        if: inputs.bump == 'major'
+        run: uv run cz bump --yes --increment MAJOR
+
+      - name: Push commit and tag
```
mikelane: MEDIUM - `git push origin main --follow-tags` hardcodes `main`. This is fine for normal use, but it creates a parity gap with the local script which uses `git push origin HEAD --follow-tags` (which pushes whatever branch is currently checked out).

The concern is operational: if someone ever triggers this workflow from a different branch (via `workflow_dispatch` on a non-main branch, which GitHub supports), it would push `main` regardless of where the workflow was triggered from, which could silently skip or stomp work.

Using `git push origin HEAD:refs/heads/${{ github.ref_name }} --follow-tags` pushes the current commit to the branch the workflow ran on, matching the dispatch context. This is what the worktree version of the file uses and it is the safer default.

scripts/release.sh
```diff
@@ -0,0 +1,29 @@
...
+# Show what's about to be released
+echo "Unreleased commits:"
+git log --oneline "$(git describe --tags --abbrev=0)"..HEAD
+echo ""
+
+# Bump version (updates pyproject.toml, __init__.py, CHANGELOG.md, creates commit + tag)
+uv run cz bump --yes $INCREMENT_FLAG
```
mikelane: LOW (SC2086) - `shellcheck` flags `$INCREMENT_FLAG` as unquoted, and it is right to do so. When `INCREMENT_FLAG` is empty, `uv run cz bump --yes $INCREMENT_FLAG` expands as expected. But bash word splitting on an unquoted variable is the mechanism making `--increment PATCH` split into two arguments — that works by accident, not by intent.

The worktree's newer version of this file fixes this properly by switching from a string variable to an array:

```bash
increment_args=()
case "${1:-auto}" in
  --patch) increment_args=(--increment PATCH) ;;
  ...
esac
uv run cz bump --yes "${increment_args[@]}"
```

This is the correct pattern for passing optional argument lists in bash. The empty array case is safe (`"${increment_args[@]}"` expands to nothing when empty), and shellcheck passes. Worth pulling that fix into this PR since the worktree already has it.

mikelane (commented):
## Review: feat(release): adopt frequent stable releases (FastAPI/Ruff model)

### Quality Checks Run

- `pre-commit run --all-files` — all 21 hooks passed (ruff, mypy, bandit, markdownlint, cspell, detect-secrets, etc.)
- `shellcheck scripts/release.sh` (against PR head commit `3658cb4`) — one finding: SC2086 on `$INCREMENT_FLAG` (see inline)
- `actionlint .github/workflows/cut-release.yml` (against PR head) — SC2086 on unquoted `$GITHUB_STEP_SUMMARY` in two steps (see inline)
- `actionlint .github/workflows/release.yml` — findings are pre-existing (SC2012 on `ls`, SC2129 on multiple redirects, outdated `softprops/action-gh-release@v1`); none introduced by this PR

---

### Summary

This is a well-motivated, clearly scoped PR. The release philosophy shift — from an RC-gated ceremony to a frequent stable releases model — is the right call for a project at this stage. The two-path approach (local script + GHA `workflow_dispatch`) is good design: it gives a solo developer the choice of terminal or browser without duplicating the underlying logic in a divergent way.

The pre-gauntlet work was thorough: the worktree version of these files is noticeably more hardened than the PR head (preflight checks, branch guard, no-commits guard, array-based args, `${{ github.ref_name }}` push). That work should come across in this PR rather than living in the worktree only.

---

### Findings

**MEDIUM — Hardcoded `main` in push step** (inline, cut-release.yml line 68)

`git push origin main --follow-tags` will always push to `main` regardless of which branch the `workflow_dispatch` was triggered on. GitHub allows dispatching workflows from any branch. Using `git push origin HEAD:refs/heads/${{ github.ref_name }} --follow-tags` matches the dispatch context and is safer. The worktree file already has this fix.

**MEDIUM — First-release guard missing from `Show unreleased commits` step** (inline, cut-release.yml line 50)

`git describe --tags --abbrev=0` exits non-zero when no tag exists. With `set -e` in the shell step, this crashes the job before bumping, with a cryptic git error. The local script guards against this with `2>/dev/null || echo ""`. The same guard is needed here. Also flags SC2086 on unquoted `$GITHUB_STEP_SUMMARY`.

**LOW — SC2086: unquoted `$INCREMENT_FLAG` in release.sh** (inline, release.sh line 22)

Word splitting on the unquoted variable is the mechanism that splits `--increment PATCH` into two arguments. It works, but by accident. The correct bash pattern is an array (`increment_args=()`). The worktree already has this fix.

**LOW — `uv sync --all-groups` installs unnecessary groups** (inline, cut-release.yml line 38)

Only `commitizen` is needed. Test, benchmark, and docs groups add setup time and supply chain surface for no benefit in a release workflow.

**LOW — `cleanup` function name is misleading** (inline, release.sh line 15)

Note: this comment applies to the worktree version only (which has the `trap` handler). The PR head version does not yet have this code. Flagging it because the worktree will likely be force-pushed or squashed into this branch.

---

### What Works Well

**The two-path design is clean.** Local script for terminal users, `workflow_dispatch` for browser users — no third path, no auto-trigger complexity. Each path is independently readable and does the same thing in the same order.

**The AGENT_GUIDELINES.md rewrite is exactly right.** Removing the RC ceremony section and replacing it with "release often, think before breaking" is pragmatic and honest about where this project actually is. The migration notes caveat for breaking changes and the major-version discussion note are the right guardrails to keep.

**OIDC trusted publishing.** The existing `release.yml` uses OIDC for PyPI publishing, which is the correct and modern approach. This PR doesn't break it.

**The step-by-step troubleshooting in the trap handler** (worktree version) is excellent observability for a script that does irreversible things. A push that fails at 2am gets a human-readable diagnosis rather than a bare exit code.

**`secrets.RELEASE_PAT` scoping note.** The PAT explanation in both the workflow and AGENT_GUIDELINES.md is clear: fine-grained, `contents: write` only, 90-day expiry. That's the right scope.

---

### Merge Recommendation

The core design is sound and the philosophy shift is well-argued. The MEDIUM findings (hardcoded `main`, missing first-release guard in the GHA step) are real operational risks worth fixing before merge — both are low-effort fixes that the worktree already has. The LOW findings are quality improvements that would bring the PR head in line with the already-improved worktree.

If the worktree's version of these files is going to be squashed in, the inline findings are addressed automatically. If this commit is being merged as-is, the MEDIUM items warrant a follow-up before the workflow is first triggered.

LGTM on architecture and philosophy. Recommend addressing the MEDIUM items (either in a fixup commit or confirming the worktree changes will be included).

PR #238, opened 2026-03-04
fix(instrumentation): insert gremlin injection after future imports at AST level

### LINKED ISSUE #237: fix(instrumentation): from __future__ import annotations causes SyntaxError in instrumented files
## Problem

`_write_instrumented_sources` in `plugin.py:716` prepends the gremlin activation injection as a raw string before the `ast.unparse` output:

```python
final_source = gremlin_active_injection + instrumented_source
```

`ast.unparse` preserves `from __future__ import annotations` as the first statement in `instrumented_source`. Prepending anything before it violates Python's requirement that `__future__` imports be the first statement in the file, causing:

```
SyntaxError: from __future__ imports must occur at the beginning of the file
```

## Affected files

All source files containing `from __future__ import annotations` — currently 41 files in `src/pytest_gremlins/`.

## Root cause

The injection is applied as text concatenation after `ast.unparse`, bypassing the AST. The fix belongs at the AST level: insert the injection nodes *after* any `from __future__` imports in `tree.body`, then unparse once.

## Proposed fix

```python
future_imports = [n for n in tree.body if isinstance(n, ast.ImportFrom) and n.module == '__future__']
rest = [n for n in tree.body if n not in future_imports]
injection_nodes = ast.parse(gremlin_active_injection).body
tree.body = future_imports + injection_nodes + rest
final_source = ast.unparse(tree)
```

## Acceptance criteria

- [ ] Files with `from __future__ import annotations` are instrumented without `SyntaxError`
- [ ] Files without `from __future__` imports continue to work
- [ ] Module docstrings (if any) are also handled correctly — they must remain before `from __future__`

mikelane:
## Summary

Fixes #237 — files containing `from __future__ import annotations` caused a `SyntaxError` during mutation testing because the gremlin injection was prepended as plain text *before* the `__future__` import, violating Python's requirement that `from __future__` must be the first non-docstring statement.

**Root cause:** `_write_instrumented_sources` prepended `__gremlin_active__ = ...` via string concatenation, placing it before any `from __future__` imports already in the source.

**Fix:** New `_prepend_injection()` helper operates at the AST level. It walks the module body, skips the module docstring (if present) and all `from __future__` imports, then inserts the injection nodes at the correct position before converting back to source via `ast.unparse`.

## Changes

| File | What changed |
|------|-------------|
| `src/pytest_gremlins/plugin.py` | Add `_prepend_injection()` helper; update `_write_instrumented_sources` to use it; rename `split`→`insert_position`, `e`→`execution_error` for clarity |
| `tests/instrumentation/test_write_instrumented_sources.py` | 10 new tests covering future import ordering, docstring handling, multiple future imports, and empty modules |
| `tests/parallel/test_pool.py` | Reclassify 5 WorkerPool subprocess tests from `small` → `medium` (enforced by pytest-test-categories 1.2.1) |
| `pyproject.toml` + `uv.lock` | Upgrade `pytest-test-categories` 1.1.0 → 1.2.1; tighten floor to `>=1.2.1` |

## Validation

Retroactive TDD check confirmed: 4 of the 7 ordering-sensitive tests failed against the pre-fix `plugin.py`. The 3 that passed on both versions test behaviors that were never broken (basic injection presence without future imports).

## Test plan

- [x] `uv run pytest tests/instrumentation/ -m small` — all 20 small instrumentation tests pass
- [x] `uv run pytest tests/parallel/ -m "small or medium"` — all 10 pool tests pass
- [x] `uv run pytest tests -m small` — full small suite passes
- [x] Pre-push hooks: ruff, mypy (strict), bandit, all pass
- [x] Church purification: naming, test coverage gaps (3 added), secret scan, arch, dep audit all pass

## Dependencies

> ⚠️ **Stacked on PR #235** (`fix/release-process`). GitHub will retarget this PR to `main` automatically once #235 merges.

---

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1419,7 +1419,7 @@ dev = [
     { name = "playwright", specifier = ">=1.58.0" },
     { name = "pre-commit", specifier = ">=3.6.0" },
     { name = "pytest-cov", specifier = ">=4.1.0" },
-    { name = "pytest-test-categories", specifier = ">=1.0.0" },
+    { name = "pytest-test-categories", specifier = ">=1.2.1" },
     { name = "pytest-xdist", specifier = ">=3.5.0" },
     { name = "ruff", specifier = ">=0.1.6" },
     { name = "tox", specifier = ">=4.11.0" },


src/pytest_gremlins/plugin.py
```diff
@@ -725,6 +726,48 @@ def _write_instrumented_sources(
...
+    """
+    insert_position = 0
+
+    if (
+        body
+        and isinstance(body[0], ast.Expr)
+        and isinstance(body[0].value, ast.Constant)
+        and isinstance(body[0].value.value, str)
+    ):
+        insert_position = 1
```
mikelane: The three-phase walk (skip docstring, skip futures, stop) is clean and easy to audit. One observation: the docstring check (lines 753–759) checks only `body[0]` via `isinstance(body[0].value.value, str)`. This correctly excludes bytes literals (`b'...'`) since `bytes` is not `str`. Python's own `ast.get_docstring()` uses the same check, so this is canonically correct behavior. No change needed — just confirming the correctness.

tests/parallel/test_pool.py
```diff
@@ -99,6 +99,7 @@ def it_submit_requires_active_context(self, tmp_path: Path) -> None:
                 env_vars={},
             )
```
mikelane: The test reclassification from `small` to `medium` is the right call. These tests spawn real subprocesses via `WorkerPool` and wait for their completion — by Google test size classification, any test that touches real processes belongs in `medium` at minimum. The comment annotations (`# Pool shutdown waits for subprocess completion`, `# Spawns real subprocess via WorkerPool`) explain the WHY for each reclassification, which is exactly the right approach for documenting non-obvious size decisions.

PR #338, opened 2026-03-26
feat(instrumentation): log exception before re-raising in import hook

mikelane:
## Summary

- Wraps the `compile()`+`exec()` in `GremlinLoader.exec_module()` with try/except that logs the exception with `exc_info=True` before re-raising
- Ensures the full traceback reaches subprocess stderr rather than being silently swallowed when instrumented code fails for complex libraries like attrs
- Adds test verifying the error is logged with module name and re-raised

Part of the effort to diagnose the 100% gremlin error rate on attrs (680/681 gremlins ERROR).

## Test plan

- [x] `uv run pytest tests/instrumentation/test_import_hooks.py -m small` — 14 passed
- [x] `uv run mypy src/pytest_gremlins` — clean
- [x] `uv run ruff check src tests` — clean
- [x] Pre-push hooks all green

### DIFF
--- a/tests/instrumentation/test_import_hooks.py
+++ b/tests/instrumentation/test_import_hooks.py
@@ -127,7 +127,7 @@ class DescribeGremlinLoader:
         record = caplog.records[0]
         assert 'bad_module' in record.message
         assert record.exc_info is not None
-        assert record.exc_info[1] is not None
+        assert isinstance(record.exc_info[1], RuntimeError)
 
     def it_exec_module_reads_active_gremlin_from_environment(self, monkeypatch):
         monkeypatch.setenv(ACTIVE_GREMLIN_ENV_VAR, 'g001')


mikelane (commented):
## Summary

Clean, focused change that adds diagnostic logging to `GremlinLoader.exec_module()`. The try/except + `logger.error(exc_info=True)` + re-raise pattern is correct and will surface full tracebacks in subprocess stderr when instrumented modules fail to execute -- exactly what is needed to diagnose the 100% gremlin error rate on attrs.

CI is green across all 18 test matrix cells (Python 3.11-3.14, macOS/Ubuntu/Windows), type checking, lint, docs, and codecov. The only failure is Security Audit, which is caused by pre-existing dependency CVEs in `pygments` (CVE-2026-4539) and `requests` (CVE-2026-25645) -- unrelated to this PR.

## Blockers

None.

## Recommendations

1. **MEDIUM** -- `tests/instrumentation/test_import_hooks.py:131-133`: The test asserts `record.exc_info[1] is not None` but does not verify the exception type. Consider tightening to `assert isinstance(record.exc_info[1], RuntimeError)`. This closes the gap between what `pytest.raises(RuntimeError, match='boom')` checks (the re-raised exception) and what the log record actually captured. Without it, a regression that logs a different exception type while still re-raising `RuntimeError` would pass silently.

2. **LOW** -- `uv.lock`: The diff includes a stowaway version sync (`1.6.0b0` -> `1.6.0`) from the `chore: sync uv.lock after version bump` commit. Not harmful since `pyproject.toml` already declares `1.6.0` on main, but it widens the PR scope beyond the stated feature. A rebase onto current main followed by `uv lock` would likely make this a no-op.

## Praise

- The `import logging` / `logger = logging.getLogger(__name__)` placement follows the exact pattern used across the rest of the codebase (10+ modules). Good consistency.
- Import ordering is correct: `logging` sits alphabetically between `importlib.machinery` and `os` in the stdlib block, and `logger` is placed between the third-party import and the `TYPE_CHECKING` block -- matching `pragma.py`, `config.py`, and other sibling modules.
- The test verifies both the logging side-effect AND the re-raise behavior in a single, focused test method. Using `caplog.at_level` scoped to the specific logger avoids cross-contamination from other log output. The parenthesized `with` block combining `caplog.at_level` and `pytest.raises` is clean.
- Bare `except Exception:` (not bare `except:`) is the right choice -- it catches all user-code exceptions without swallowing `SystemExit` or `KeyboardInterrupt`.
- The test inherits `@pytest.mark.small` from the class decorator and follows the `it_*` naming convention with a descriptive, non-"should" name.

## Decision

**APPROVED** -- The recommendations above are non-blocking improvements. The core change is correct, well-tested, and consistent with project conventions.

PR #344, opened 2026-03-27
perf: lightweight test runner + batch early termination fix

mikelane:
## Summary

- **fix(batch):** Remove early termination in `_run_gremlin_batch()` that silently skipped gremlins after the first zap. Batch mode was testing only 17/117 gremlins, producing incorrect mutation scores. Now all gremlins in each batch are tested independently.

- **perf:** Add a lightweight test runner that bypasses full pytest startup for per-gremlin subprocesses. Each subprocess was spending ~900ms on pytest framework overhead (plugin loading, test collection, fixture setup) for ~50ms of actual test execution. The lightweight runner imports test modules directly and calls test functions, reducing per-gremlin cost from 949ms to 197ms.

### Benchmark results (synthetic project: 117 gremlins, 55 tests)

| Mode       | Before  | After  | Speedup |
|------------|---------|--------|---------|
| Sequential | 111.0s  | 22.5s  | **4.9x** |
| Parallel   | 17.2s   | 5.1s   | **3.4x** |
| Batch      | broken  | 5.7s   | **fixed** |

### Benchmark results (attrs: 681 gremlins, 1337 tests)

| Metric | Value |
|--------|-------|
| Wall time (parallel) | **86.5s** (was ~315s baseline, **3.6x faster**) |
| Gremlins zapped | 662/681 (97%) |

Mutation detection quality unchanged — same gremlins killed/survived in all modes.

### How the lightweight runner works

1. Reads `pyproject.toml` for `pythonpath` config (like pytest does)
2. Sets up the same import hooks as the bootstrap script (reads `sources.json`)
3. Imports test modules directly via `importlib.util`
4. Calls test functions/methods and checks for exceptions
5. Falls back to full pytest when the runner script is unavailable

## Test plan

- [x] All 1311 existing tests pass
- [x] Batch executor tests updated for new behavior (no early termination)
- [x] Synthetic benchmark: 117 gremlins, 91 killed (77.8%) — matches baseline exactly
- [x] attrs benchmark: 681 gremlins, 662 killed (97%)
- [x] mypy strict passes
- [x] ruff check + format passes

### DIFF
--- a/src/pytest_gremlins/parallel/persistent_pool.py
+++ b/src/pytest_gremlins/parallel/persistent_pool.py
@@ -31,7 +31,6 @@ from concurrent.futures import (
 )
 import logging
 import os
-from pathlib import Path
 import subprocess
 import time
 from typing import (
@@ -42,6 +41,7 @@ from typing import (
 if TYPE_CHECKING:
     import multiprocessing
 
+from pytest_gremlins.parallel.lightweight import build_lightweight_command
 from pytest_gremlins.parallel.pool import WorkerResult
 from pytest_gremlins.parallel.pool_config import PoolConfig
 from pytest_gremlins.reporting.results import GremlinResultStatus
@@ -59,38 +59,6 @@ def _warmup_noop() -> bool:  # pragma: no cover
         True to indicate successful warmup.
     """
     return True
-
-
-def _build_lightweight_command(
-    test_command: list[str],
-    env_vars: dict[str, str],
-) -> list[str] | None:
-    """Build a lightweight runner command if the runner script exists.
-
-    Extracts test node IDs from the full test command and builds a
-    command using the lightweight runner (no pytest overhead).
-
-    Args:
-        test_command: Original test command (e.g. [python, bootstrap.py, -x, ...]).
-        env_vars: Environment variables that may contain sources file path.
-
-    Returns:
-        Lightweight command list, or None if the runner is not available.
-    """
-    sources_file = env_vars.get('PYTEST_GREMLINS_SOURCES_FILE', '')
-    if not sources_file:
-        return None
-
-    runner_path = Path(sources_file).parent / 'gremlin_lightweight_runner.py'
-    if not runner_path.exists():
-        return None
-
-    # Extract test node IDs from test_command (args containing '::')
-    test_ids = [arg for arg in test_command[2:] if '::' in arg]
-    if not test_ids:
-        return None
-
-    return [test_command[0], str(runner_path), *test_ids]
 
 
 def _run_gremlin_batch(  # pragma: no cover
@@ -116,7 +84,7 @@ def _run_gremlin_batch(  # pragma: no cover
     Returns:
         List of WorkerResult for each tested gremlin.
     """
-    lightweight_cmd = _build_lightweight_command(test_command, env_vars)
+    lightweight_cmd = build_lightweight_command(test_command, env_vars)
     effective_command = lightweight_cmd if lightweight_cmd is not None else test_command
 
     results: list[WorkerResult] = []
@@ -141,8 +109,20 @@ def _run_gremlin_batch(  # pragma: no cover
 
             execution_time_ms = (time.monotonic() - start_time) * 1000
 
-            if result.returncode != 0:
-                # Mutation caught - test failed
+            # Only return code 1 (tests failed) should be treated as a mutation
+            # being zapped. Other non-zero return codes indicate pytest errors
+            # (collection/import/internal) and should not inflate the score.
+            if result.returncode == 0:
+                # Mutation survived - tests passed
+                results.append(
+                    WorkerResult(
+                        gremlin_id=gremlin_id,
+                        status=GremlinResultStatus.SURVIVED,
+                        execution_time_ms=execution_time_ms,
+                    )
+                )
+            elif result.returncode == 1:
+                # Mutation caught - tests failed
                 results.append(
                     WorkerResult(
                         gremlin_id=gremlin_id,
@@ -152,12 +132,16 @@ def _run_gremlin_batch(  # pragma: no cover
                     )
                 )
             else:
-                # Mutation survived - test passed
+                # Other non-zero exit codes indicate errors
+                error_output = ''
+                if result.stderr:
+                    error_output = result.stderr.decode(errors='replace')[:2000]
                 results.append(
                     WorkerResult(
                         gremlin_id=gremlin_id,
-                        status=GremlinResultStatus.SURVIVED,
+                        status=GremlinResultStatus.ERROR,
                         execution_time_ms=execution_time_ms,
+                        error_output=error_output,
                     )
                 )
         except subprocess.TimeoutExpired:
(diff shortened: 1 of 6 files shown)


src/pytest_gremlins/parallel/persistent_pool.py
```diff
@@ -60,18 +61,50 @@ def _warmup_noop() -> bool:  # pragma: no cover
...
+    runner_path = Path(sources_file).parent / 'gremlin_lightweight_runner.py'
+    if not runner_path.exists():
+        return None
+
+    # Extract test node IDs from test_command (args containing '::')
+    test_ids = [arg for arg in test_command[2:] if '::' in arg]
+    if not test_ids:
+        return None
+
+    return [test_command[0], str(runner_path), *test_ids]
```
mikelane: **CRITICAL -- CI coverage gate failure (96% < 97% fail-under)**

All 8 Linux/macOS test jobs fail because this new helper has zero test coverage. `_build_lightweight_command` is a pure function and trivially unit-testable. Parametrized cases needed:
1. Returns `None` when `PYTEST_GREMLINS_SOURCES_FILE` is absent
2. Returns `None` when runner script does not exist on disk
3. Returns `None` when no test IDs contain `::`
4. Returns the lightweight command list when all preconditions are met

The inline duplication in `plugin.py:2403-2405` also needs coverage.

Merge blocker -- CI must be green.

src/pytest_gremlins/plugin.py
```diff
@@ -1273,6 +1276,170 @@ def find_spec(self, fullname, path, target=None):
...
+            func = getattr(module, parts[1], None)
+            if func is None:
+                return True
+            func()
+        else:
+            return True
+
+        return True
+    except (AssertionError, Exception):
+        return False
```
mikelane: **HIGH -- `return True` on module/class/method not found lets mutations silently survive**

Every lookup failure path (lines 1363, 1368, 1372, 1377, 1380) returns `True` (test passed, mutation survived). This is the wrong safe default for a mutation tester.

If a test module fails to import (e.g. the lightweight runner does not support conftest fixtures, or a path mismatch occurs), the gremlin is reported as SURVIVED rather than ERROR. This inflates the mutation score and hides real problems.

The safe default is `False`. A mutation tester must be pessimistic -- "I could not verify this mutation was caught" should not count as "survived."

Also on line 1383: `except (AssertionError, Exception)` is redundant since `AssertionError` is a subclass of `Exception`. Simplify to `except Exception`.

[BUG-HUNTER HIGH]

src/pytest_gremlins/parallel/persistent_pool.py
```diff
@@ -91,10 +127,11 @@ def _run_gremlin_batch(  # pragma: no cover
...
         env['ACTIVE_GREMLIN'] = gremlin_id
+        env['GREMLIN_ROOTDIR'] = rootdir
 
         try:
             result = subprocess.run(  # Intentional: runs pytest test commands
-                test_command,
+                effective_command,
                 cwd=rootdir,
                 env=env,
                 capture_output=True,
```
mikelane: **HIGH -- `returncode != 0` treats all non-zero exits as ZAPPED, unlike `pool.py` which correctly uses `returncode == 1`**

`pool.py:_run_gremlin_test` (line 108-129) correctly distinguishes:
- `returncode == 0` -> SURVIVED
- `returncode == 1` -> ZAPPED
- other non-zero -> ERROR

But this batch path uses `returncode != 0` (line 144), treating import errors (exit 2), internal pytest errors (exit 3+), and lightweight runner crashes as "mutation caught." This inflates the mutation score.

The lightweight runner makes this worse: if it crashes for any reason, the exit code is non-zero and the batch path incorrectly counts that as ZAPPED.

Consider aligning: only `returncode == 1` is ZAPPED, other non-zero codes are ERROR.

tests/parallel/test_batch_executor.py
```diff
@@ -127,12 +127,14 @@ def it_terminates_batch_early_when_configured(self, tmp_path: Path) -> None:
...
-        # g001 survives, g002 is zapped, g003 is skipped (within same batch)
-        assert len(results) == 2
+        # All gremlins tested: g001 survives, g002 zapped, g003 survives
+        assert len(results) == 3
         assert results[0].gremlin_id == 'g001'
         assert results[0].status == GremlinResultStatus.SURVIVED
         assert results[1].gremlin_id == 'g002'
         assert results[1].status == GremlinResultStatus.ZAPPED
+        assert results[2].gremlin_id == 'g003'
+        assert results[2].status == GremlinResultStatus.SURVIVED
```
mikelane: **MEDIUM -- Test name `it_terminates_batch_early_when_configured` contradicts the tested behavior**

This test now asserts all 3 gremlins are tested (exhaustive execution). The name claims early termination. Rename to something like `it_tests_all_gremlins_in_batch_regardless_of_zap`.

[TEST MEDIUM]

tests/parallel/test_persistent_pool.py
```diff
@@ -332,12 +332,14 @@ def it_submit_batch_stops_on_first_failure(self, tmp_path: Path) -> None:
...
-            # g001 survives, g002 is zapped (caught), g003 should be skipped
-            assert len(results) == 2
+            # All gremlins tested: g001 survives, g002 zapped, g003 survives
+            assert len(results) == 3
             assert results[0].gremlin_id == 'g001'
             assert results[0].status == GremlinResultStatus.SURVIVED
             assert results[1].gremlin_id == 'g002'
             assert results[1].status == GremlinResultStatus.ZAPPED
+            assert results[2].gremlin_id == 'g003'
+            assert results[2].status == GremlinResultStatus.SURVIVED
```
mikelane: **MEDIUM -- Test name `it_submit_batch_stops_on_first_failure` contradicts the tested behavior**

Assertions verify exhaustive execution (3 results returned), but the name and docstring (line 311) claim stop-on-first-failure. Rename to `it_submit_batch_tests_all_gremlins_even_after_zap` and update the docstring.

[TEST MEDIUM]

src/pytest_gremlins/parallel/pool.py
```diff
@@ -78,10 +78,21 @@ def _run_gremlin_test(  # pragma: no cover
...
+
+    # Use lightweight runner if available (skips full pytest startup)
+    effective_command = test_command
+    sources_file = env_vars.get('PYTEST_GREMLINS_SOURCES_FILE', '')
+    if sources_file:
+        runner_path = Path(sources_file).parent / 'gremlin_lightweight_runner.py'
+        if runner_path.exists():
+            test_ids = [arg for arg in test_command[2:] if '::' in arg]
+            if test_ids:
+                effective_command = [test_command[0], str(runner_path), *test_ids]
```
mikelane: **LOW -- Lightweight command construction duplicated in 3 places**

This logic duplicates `_build_lightweight_command` from `persistent_pool.py`. The same pattern also appears in `plugin.py:2398-2405`. All three sites should call the extracted helper to avoid divergence and simplify future changes (e.g., the returncode alignment fix would only need to happen in one place).

[ARCH LOW]

PR #398, opened 2026-04-24
feat: add --gremlin-explain diagnostic flag (#397)

### LINKED ISSUE #387: bug: coverage-guided selection drops killing tests, producing false-positive survivors (1.8.1)
## Summary

In 1.8.1, coverage-guided test selection sometimes excludes tests that would kill a given mutant, producing false-positive survivors. Adding `--gremlin-no-coverage-filter` makes the same tests kill the same mutants, so the test suite is adequate and the failure is in the selection layer.

Same neighborhood as the recent #356 / #367 / #368 / #376 / #377 fixes, but distinct from each: this reproduction has no duplicate function names and the coverage-stopped path is not involved.

## Version

- `pytest-gremlins==1.8.1` (latest on PyPI as of report date)
- `pytest==9.0.2`
- `coverage==7.x` (whatever the resolver picked alongside 1.8.1)
- `python==3.12`
- macOS arm64 + Linux x86_64 self-hosted runners — symptom reproduces on both

## Reproduction

Reproduced on a feature branch of `SayMoreAI/saymore` (private repo, ~1100 tests across `tests/test_synthesis_*` and `tests/test_*_port.py`). The bug is reliable run-to-run, but the specific surviving mutants shift, which points at the cache+selection interaction rather than a deterministic per-mutant flaw.

Broken run:

```bash
uv run pytest --gremlins --gremlin-cache --gremlin-workers=auto
# Result: Zapped: 152 (46%), Survived: 67 (20%), Pardoned: 114
```

Diagnostic step that proved the survivors are false positives. Pick one survivor — e.g. a `==` → `!=` mutation in a routing function — and apply it manually:

```python
# saymore/synthesis/tools.py:46
# Original:
if tool_name == 'get_conversations':
# Mutated to:
if tool_name != 'get_conversations':
```

Then run **only** the tests that exercise that file:

```bash
uv run pytest tests/test_synthesis_tools.py
# Result: 10 tests fail immediately with assertion errors
```

So the test suite kills that mutant when it gets to run those tests. The selection layer was excluding all 10 of them.

Fixed run (with workaround):

```bash
uv run pytest --gremlins --gremlin-clear-cache --gremlin-workers=auto --gremlin-no-coverage-filter
# Result: Zapped: 219 (66%), Survived: 0 (0%), Pardoned: 114
```

Same test suite, same mutants, zero survivors. The 67 survivors from the first run all get killed once the coverage filter is bypassed.

## Expected behavior

For a mutant in source line `L`, the selection layer should run **every** test that has `L` in its coverage data — at minimum, every test in the test file that imports the module containing `L`. With this guarantee, a test that fails on the mutated line will always be selected to attempt the kill.

## Actual behavior

In 1.8.1, some mutants are run against a strictly smaller set of tests than the set that has `L` in their coverage. The dropped tests are the ones that would have killed the mutant. Result: false-positive "survivor" reported, even though the test suite as a whole is adequate.

## What it isn't

- A missing-test problem. Manually applying the mutation and running the relevant test file kills it on the first try.
- A `# gremlin: pardon[*]` issue. None of the survivors carry pardon pragmas; they were scored as real survivors.
- An incompletely-shipped fix from #356/#367/#368/#376/#377. Those PRs are in 1.8.1; the symptom still reproduces.
- An xdist/parallelism artifact. We use `--gremlin-workers=auto` (gremlin's own pool, not pytest-xdist), and disabling parallelism doesn't change the survivor set.

## Hypotheses (in order of suspicion)

1. **Cache key staleness** when coverage data changes but source doesn't. We use `--gremlin-cache` for incremental runs. If the cache key is hashed only over source content but the *coverage→test* mapping is computed from a stale coverage database, the cache can hand back a selection set that excludes valid killing tests on the next run. Reproduction supports this: removing `--gremlin-cache` (`--gremlin-clear-cache`) plus `--gremlin-no-coverage-filter` returns clean; we have not yet isolated whether `--gremlin-clear-cache` alone (with the filter still on) is enough.
2. **Coverage subprocess granularity vs. mutation site granularity**. Even with full node IDs (#368), if the coverage subprocess records line ranges that don't include the exact mutated line for tests that *do* exercise it (e.g. due to branch-coverage gap, conditional import, or module-level code), the selection layer will skip those tests. The mutations we saw survive were inside small, hot functions exercised by many tests — exactly where this kind of off-by-one would matter least, which makes it suspicious.
3. **Filter excludes tests that the mutant transitively reaches via tools/imports** rather than via direct file coverage. e.g. a test in `test_synthesis_graph.py` that drives `_execute_tools` which imports and calls `tools.execute_tool` — if the filter only checks "does this test have a coverage hit on tools.py?" rather than "does this test invoke any function that reaches tools.py?", it could miss those tests for a `tools.py` mutation.

Low confidence on which is the root cause. (1) and (2) match the recent fix history most closely. I can instrument a specific hypothesis if you want to narrow it down.

## Workaround

Use `--gremlin-no-coverage-filter` in CI. We landed this in `.github/workflows/ci.yml`:

```yaml
- name: Run mutation tests
  run: uv run pytest --gremlins --gremlin-workers=auto --gremlin-cache --gremlin-no-coverage-filter
```

After applying the workaround, our mutation gate goes from `Survived: 67` → `Survived: 0` with no test additions, no pragma additions, no source changes. The false positives extended into pre-existing files the branch did not touch (`saymore/conversation/graph.py`, `saymore/conversation/prompt.py`, `saymore/adapters/*`, `saymore/synthesis/privacy.py`) — every one of them dropped to zero survivors after adding the flag. The bug is in the selection layer, not in any specific test or mutation.

## Suggested follow-ups

1. **Diagnostic flag** — add a `--gremlin-debug-coverage-selection` flag (or `--gremlin-explain <node-id>`) that, for a given surviving mutant, prints the set of tests that *had coverage* on the mutated line vs. the set that was *actually selected*. The delta is the bug.
2. **Cache invalidation** — if (1) above turns out to be the root cause, the cache key should incorporate the selection-layer's view of the coverage→test map, not just source hashes. Or `--gremlin-cache` should detect coverage-database staleness and warn/invalidate.
3. **Defensive default** — consider making `no_coverage_filter` the default for the next minor release and re-adding the filter as an explicit opt-in (`--gremlin-fast-coverage-filter`?) once the selection bug is closed. The cost of extra test runs is much lower than the cost of false-positive survivors silently lying about test adequacy.
4. **Documentation** — until this is closed, the README should mention `--gremlin-no-coverage-filter` as the recommended flag for "I want correct results, not fast results."

## Related issues

- #356, #367, #368, #376, #377 — same surface area, recently fixed
- #350 — coverage-selection debuggability (the `selected_tests` field in JSON reports would have helped me triage this faster; it would also help any user hit by this bug)

## Available on request

- A minimal reproduction repository if hypothesis (1) doesn't isolate locally
- The `gremlins.json` from a broken run (sanitized)
- The exact source-coverage delta for one of the false-positive survivors

### LINKED ISSUE #397: feat: --gremlin-explain flag for diagnosing coverage-selection drops (supports #387)
## Summary

Add a read-only `--gremlin-explain=<gremlin_id>` diagnostic flag that, for one
gremlin, prints the diff between "tests whose coverage map contains the mutated
line" and "tests actually selected by `_select_tests_for_gremlin`". This
surfaces node-ID drift, selection gaps, and other selection-layer bugs at a
single glance without running mutation testing.

## Motivation

This is #387's suggested follow-up #1. The diagnostic is valuable on its own
(ongoing debuggability for coverage-selection bugs), but it also unblocks the
#387 root-cause fix by providing empirical evidence from the reporter's own
repo (SayMoreAI/saymore) before we commit to a fix approach.

## Acceptance criteria

- [ ] CLI flag `--gremlin-explain=<gremlin_id>` added near other `--gremlin-*`
      flags in the argparse group; stored on `GremlinSession` as
      `explain_gremlin_id: str | None`.
- [ ] When set, the plugin prints a diagnostic for the named gremlin — the
      covering set, the selected list, the diff, and side-by-side normalized
      forms that surface marker/path/casing drift — then disables the gremlin
      session (without `sys.exit` or `pytest.exit`) so the rest of pytest can
      finish cleanly.
- [ ] Uses `difflib.get_close_matches(raw, candidates, n=3, cutoff=0.7)` to
      suggest likely-intended node IDs for any dropped tests.
- [ ] Unit test in `tests/plugin/test_gremlin_explain.py` with a hand-built
      `CoverageMap` that simulates a one-char drift.
- [ ] Pytester integration test in
      `tests/plugin/test_gremlin_explain_pytester.py` that uses an inline
      plugin to simulate a `[fixture-1]` suffix and asserts the diagnostic
      names the dropped test.
- [ ] Flag does NOT modify coverage collection, selection behavior, or the
      persistent JSON report schema.

## Non-goals

- Does not fix #387. A separate PR (after this one merges and the diagnostic
  is validated against the reporter's repro) introduces the canonical-key
  normalization that fixes the underlying bug.
- Does not flip any defaults or change existing flags.

## Related

- Supports #387 (empirical validation step)
- Implementation plan: `/Users/mikelane/.claude/plans/make-a-plan-to-rustling-honey.md` (local to reporter)

🤖 Planned with [Claude Code](https://claude.com/claude-code)

mikelane:
## Summary

Adds a read-only `--gremlin-explain=<gremlin_id>` CLI flag that, for a single
gremlin, prints the diff between "tests whose coverage map contains the
mutated line" and "tests actually selected by the prioritized selector."
The output surfaces node-ID drift, missing runnable mappings, and close-match
suggestions — exactly the debuggability the reporter asked for in #387.

## Why this PR is useful on its own

This is **PR 1 of a two-PR sequence**. PR 2 (still to open) will apply the
canonical-node-ID-key fix that actually closes #387. Shipping the diagnostic
first lets the reporter run `--gremlin-explain=<one-survivor-id>` against
their private repro (SayMoreAI/saymore) and **empirically confirm the
root-cause hypothesis** — asymmetric marker stripping between
`_make_node_ids_relative` (main session; `[A-Z]+` regex) and
`_strip_nodeid_markers` (coverage subprocess; any `' ['`) — before we
commit to the fix approach in PR 2.

Also valuable long-term: whenever coverage-selection disagreements surface
in the future, users have a first-class tool to introspect them instead of
filing \"this mutation shouldn't have survived\" bugs with no repro.

## What changed

- **New CLI flag** `--gremlin-explain=<gremlin_id>` wired into the argparse
  group next to the existing `--gremlin-*` flags.
- **New `GremlinSession.explain_gremlin_id: str | None`** field captured in
  `pytest_configure` from the CLI option.
- **New short-circuit hook** in `pytest_sessionfinish` that, when the flag
  is set, prints the diagnostic and disables the gremlin session so the
  mutation loop no-ops without calling `sys.exit` or `pytest.exit`. Pytest
  finishes its own session cleanly and user tests still count.
- **Nine new private helpers** decompose the explainer into single-purpose
  functions (`_emit_selection_explainer`, `_maybe_short_circuit_for_explain`,
  `_dispatch_mutation_run`, `_covering_tests_for_gremlin`,
  `_print_explainer_header`, `_close_matches_display`, `_print_dropped_tests`,
  `_print_unrunnable_selections`, `_drop_bracketed_suffix`).
- **12 new tests** — 10 `small` unit tests + 2 `medium` pytester end-to-end
  tests — covering every branch: drift present, no drift, unknown gremlin id,
  flag absent, banner/counts/close-match content, short-circuit semantics,
  exit code, and the exact \"no gremlin with id ...\" phrasing.
- **`difflib`** is the only new import; stdlib; no supply-chain surface.
- **`uv.lock`** one-line sync to version 1.8.1 (unrelated, needed to clear
  the pre-push hook blocker from commit 0bfd60c).

## What this PR is NOT

- It does **not** fix #387. No changes to `_make_node_ids_relative`, the
  `[A-Z]+` regex, coverage collection, selection behavior, or the persistent
  JSON report schema. PR 2 owns all of that.
- It does **not** flip any existing defaults. `--gremlin-no-coverage-filter`
  behavior is unchanged.

## Test plan

- [x] Unit tests pass: `uv run pytest tests/plugin/test_gremlin_explain.py -v`
  (10 tests)
- [x] Pytester integration tests pass:
  `uv run pytest tests/plugin/test_gremlin_explain_pytester.py -v` (2 tests)
- [x] Broader small+medium suite green:
  `uv run pytest -m \"small or medium\"` (no regressions introduced by this PR)
- [x] Ruff/format/mypy clean across `src/` and `tests/`
- [ ] Reporter (solo dev — me) runs
  `uv run pytest --gremlins --gremlin-explain=<survivor_id>` against
  SayMoreAI/saymore post-merge to empirically validate the node-ID-drift
  hypothesis before PR 2 is opened

## Pre-review gauntlet

Full `/preparing-for-review` gauntlet was run before this PR:

| Phase | Status | Notes |
|---|---|---|
| Bug hunting | PASS | 0 confirmed bugs |
| Quality audit | PASS | 1 addressed — assertion-tightening (commit ada6ed4) |
| Security audit | PASS | 0 findings (1 INFO, non-actionable) |
| Church purification (14 parallel agents) | PASS | 3 made improvements (commit b960170); others N/A or clean |

## Commits

1. `0b58a1e feat: add --gremlin-explain diagnostic flag (#397)` — implementation + tests
2. `faf5994 chore: sync uv.lock with pyproject.toml version bump to 1.8.1` — unrelated lockfile sync
3. `ada6ed4 test: tighten --gremlin-explain drift-detection assertions (#397)` — structured behavioral checks
4. `b960170 refactor: apply church-sweep polish for --gremlin-explain (#397)` — arch/test/copy purist improvements

## Links

- Closes #397
- Supports #387 (the PR that will close #387 comes next; this one unblocks it)

### DIFF
--- a/src/pytest_gremlins/plugin.py
+++ b/src/pytest_gremlins/plugin.py
@@ -548,9 +548,10 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=None,
         dest='gremlin_explain',
         help=(
-            'Print a diagnostic for the given gremlin id showing the covering set, '
-            'the selected list, their diff, and close-match suggestions for dropped '
-            'tests; then exit without running mutation testing.'
+            'Diagnose why a gremlin survived: for the given gremlin id (e.g. g001), '
+            'print the covering set, the selected list, their diff, and close-match '
+            'suggestions for any dropped tests, then skip the mutation loop so pytest '
+            'can finish normally.'
         ),
     )
 
@@ -2215,7 +2216,11 @@ def _emit_selection_explainer(gremlin_session: GremlinSession) -> None:
     )
 
     if target_gremlin is None:
-        print(f'--gremlin-explain: no gremlin with id {target_id!r} in this session.')
+        print(
+            f'--gremlin-explain: no gremlin with id {target_id!r} in this session. '
+            f'Run with --gremlins (without --gremlin-explain) to see the full list of '
+            f'gremlin ids in the progress log.'
+        )
         gremlin_session.enabled = False
         return
 
@@ -2228,7 +2233,11 @@ def _emit_selection_explainer(gremlin_session: GremlinSession) -> None:
     _print_explainer_header(target_gremlin, covering, selected)
 
     if not covering_minus_selected and not selected_minus_runnable:
-        print('  Result: selection is consistent with covering set (no drift).')
+        print(
+            '  Result: selection is consistent with covering set (no drift). '
+            'If this gremlin still survived, the cause is elsewhere (e.g. a weak '
+            'assertion in the covering tests).'
+        )
         gremlin_session.enabled = False
         return
 
--- a/src/pytest_gremlins/plugin.py
+++ b/src/pytest_gremlins/plugin.py
@@ -16,6 +16,7 @@ from dataclasses import (
     field,
 )
 from dataclasses import replace as dataclass_replace
+import difflib
 from enum import Enum
 import functools
 import importlib.util
@@ -231,6 +232,7 @@ class GremlinSession:
     max_pardons: int | None = None
     no_coverage_filter: bool = False
     test_name_to_node_ids: dict[str, list[str]] = field(default_factory=dict)
+    explain_gremlin_id: str | None = None
 
 
 _gremlin_session: GremlinSession | None = None
@@ -540,6 +542,17 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         dest='gremlin_no_coverage_filter',
         help='Disable coverage-guided test selection (slow but useful for debugging).',
     )
+    group.addoption(
+        '--gremlin-explain',
+        action='store',
+        default=None,
+        dest='gremlin_explain',
+        help=(
+            'Print a diagnostic for the given gremlin id showing the covering set, '
+            'the selected list, their diff, and close-match suggestions for dropped '
+            'tests; then exit without running mutation testing.'
+        ),
+    )
 
 
 def _init_cache(
@@ -728,6 +741,7 @@ def pytest_configure(config: pytest.Config) -> None:
             max_pardons_pct=toml_max_pardons_pct,
             max_pardons=toml_max_pardons,
             no_coverage_filter=bool(getattr(config.option, 'gremlin_no_coverage_filter', False)),
+            explain_gremlin_id=getattr(config.option, 'gremlin_explain', None),
             xdist_active=xdist_active,
             xdist_workers=xdist_worker_int if xdist_active else None,
         )
@@ -1569,14 +1583,36 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:  # n
     if cov_plugin is not None and hasattr(cov_plugin, 'cov') and cov_plugin.cov is not None:
         cov_plugin.cov.load()
 
-    # Choose execution mode based on configuration
+    if _maybe_short_circuit_for_explain(gremlin_session):
+        return
+
+    gremlin_session.results = _dispatch_mutation_run(session, gremlin_session)
+
+
+def _maybe_short_circuit_for_explain(gremlin_session: GremlinSession) -> bool:
+    """Run ``--gremlin-explain`` if requested and report whether to bail out.
+
+    When ``explain_gremlin_id`` is set we print the diagnostic, disable the
+    session, and tell the caller to skip the mutation loop. Returns ``True``
+    when the caller must return without running mutations.
+    """
+    if gremlin_session.explain_gremlin_id is None:
+        return False
+    _emit_selection_explainer(gremlin_session)
+    gremlin_session.results = []
+    return True
+
+
+def _dispatch_mutation_run(
+    session: pytest.Session,
+    gremlin_session: GremlinSession,
+) -> list[GremlinResult]:
+    """Pick the execution mode (batch / parallel / default) and run it."""
     if gremlin_session.batch_enabled:
-        results = _run_batch_mutation_testing(session, gremlin_session)
-    elif gremlin_session.parallel_enabled:
-        results = _run_parallel_mutation_testing(session, gremlin_session)
-    else:
-        results = _run_mutation_testing(session, gremlin_session)
-    gremlin_session.results = results
+        return _run_batch_mutation_testing(session, gremlin_session)
+    if gremlin_session.parallel_enabled:
+        return _run_parallel_mutation_testing(session, gremlin_session)
+    return _run_mutation_testing(session, gremlin_session)
 
 
 def _make_node_ids_relative(node_ids: list[str], rootdir: Path) -> list[str]:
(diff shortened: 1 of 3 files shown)


mikelane (commented):
## Review: `--gremlin-explain` diagnostic flag (#397)

**Decision: APPROVE**

### Summary

This PR adds a read-only diagnostic CLI flag that surfaces coverage-guided selection drift for a single gremlin — exactly the tool needed to diagnose the #387 class of bugs. The implementation is sound, the diagnostic is honest, and the tests are contract-proving.

### Phase 0 — Bug Verification

No bugs found. The adversarial QA session (e2ecbe9, 0 findings) is authoritative and corroborated by this review.

### Phase 1 — Quality Tools

CI: 21/21 green. codecov/patch passing. mergeStateStatus: CLEAN. No re-run needed.

### Phase 2 — Config Integrity

No changes to `.pre-commit-config.yaml`, `pyproject.toml`, or checker configs. Clean.

### Phase 3 — Code Standards

`import difflib` added to stdlib section. All new public helpers carry full type hints. No bare `except:`, no wildcard imports.

### Phase 4 — Self-Documenting Code

The `_emit_selection_explainer` docstring is notably good: it explains *why* the function is side-effect-heavy (emits to stdout, disables session) and *why* it avoids `sys.exit` (preserves the user's regular test results). These are exactly the WHY comments that belong in a codebase. The reasoning is present inline where it matters.

### Phase 5 — Test Quality

Strong. Two test files cover distinct layers:

- `test_gremlin_explain.py`: 380 lines of `@pytest.mark.small` unit tests exercising every branch through the public API — drift path, no-drift path, unknown-id path, no-op path, and each helper independently. The `_build_session_with_drift` fixture precisely models the #387 shape. Assertions are exact (equality on repr-quoted strings, not just substring membership). No loops, no private-method access, no `assert True`.
- `test_gremlin_explain_pytester.py`: Two `@pytest.mark.medium` end-to-end tests that run the real plugin via `pytester`, verify stdout structure, and confirm the mutation loop is short-circuited. The `_discover_gremlin_id` helper ties the test to user-visible output rather than internal file layouts — a deliberate and correct design choice, documented with a comment.

The `create_autospec(PrioritizedSelector, instance=True)` mock discipline is correct: no bare `MagicMock()` on the selector.

### Phase 6 — Architecture

The `_dispatch_mutation_run` extraction is a clean refactor: the `elif` chain becomes a guard-clause sequence and the session assignment moves to the caller. No net complexity added. `_maybe_short_circuit_for_explain` is a single-responsibility predicate. All helpers are under 15 lines.

**Critical honesty verification**: The explainer at line 2228 calls `_select_tests_for_gremlin_prioritized` — the exact same function real execution calls at lines 1906, 2032, and 2358. The diagnostic cannot lie about what would be selected; it uses the live selection path.

### Phase 7 — Security

No user-controlled input reaches shell commands or file writes. The `--gremlin-explain` value is a gremlin ID string used only as a lookup key in a list comprehension. No vulnerabilities.

### Phase 8 — Regression Check

Merge-base IS origin/main HEAD (zero files changed on main since divergence). No squash-merge hazard. The 7 deletions are the inlined `pytest_sessionfinish` dispatch block extracted into `_dispatch_mutation_run` — a clean mechanical extraction with no behavioral change.

### Deferred (LOW, follow-up issue to be filed)

1. **LOW**: Under `--gremlin-no-coverage-filter`, the covering-count line reads "0 test(s)" because `coverage_collector` is `None` in that mode. Harmless — `_covering_tests_for_gremlin` documents and handles the `None` case explicitly. Could be polished with a mode-aware message.
2. **LOW**: Zero-gremlin session + `--gremlin-explain` prints the unknown-id message rather than a mode-specific "no gremlins found" notice. Technically accurate but may confuse users. A mode-aware guard would improve UX.

Both findings are pre-acknowledged by the adversarial QA session. Suitable for a single follow-up polish issue.

PR #423, opened 2026-06-04
cleanup: scope cold-run coverage to gremlin source files

### LINKED ISSUE #422: cleanup: scope cold-run coverage to gremlin source files (drop wasted whole-tree tracing)
## Context

Discovered while investigating cold-cache perf (#336). The cold coverage run in `plugin.py` writes a coveragerc with `source = .`, so coverage.py traces the **entire tree** — all source AND all test files. But `_collect_coverage` already builds `gremlin_paths_map` and **discards coverage for every file that isn't a gremlin source file**. So measuring non-gremlin files is pure wasted work that gets thrown away immediately afterward.

## Change

Scope coverage.py's measurement to just the gremlin source files up front (`[run] include = <resolved gremlin file paths>`) instead of `source = .`. The resulting `coverage_by_test` map is **identical** — this pushes a filter that already exists in post-processing up into coverage's config so the work is never done.

## Honest framing — this is cleanup, NOT a perf win

Benchmarked on attrs 25.3.0: traced files 51→13, coverage DB ~25% smaller, but **end-to-end wall time only ~1%** (within noise, below the 2% bar). The cold-coverage step is dominated by **executing the test bodies**, not by tracing/SQLite overhead on non-gremlin files. So this ships as a tidy-up (less I/O, scopes measurement to what's actually used), explicitly **not** as a performance feature. The real cold-cache lever (reducing which tests run during the coverage pass) is out of scope and tracked separately under #336.

## Correctness

Critical: must NOT reintroduce false survivors (#387 class). Absolute resolved include paths match coverage.py's recorded paths; verified identical mutation outcomes (zapped/survived counts byte-identical) in A/B. Implemented + tested on branch `spike-coverage-scope` (458f01c).

mikelane:
Closes #422.

Scopes the cold coverage run to gremlin source files (`[run] include = <resolved gremlin paths>`) instead of `source = .`, since the post-processing already discards coverage for every non-gremlin file. **Ships as cleanup, not a perf win** — benchmarked ~1% on attrs (within noise; the cold-coverage step is dominated by test-body execution, not tracing). Traced files 51→13, coverage DB ~25% smaller.

## Correctness (the important part)
Adversarial QA caught that coverage.py treats `include =` entries as **fnmatch globs**, so a gremlin path with `*?[]` (e.g. `mod[v2]/`) would fail to match → false survivor (#387 class). Fix: fall back to `source = .` whenever any gremlin path contains a glob-special char; use scoped `include =` only for glob-clean paths. The predicate `*?[]` was verified an exhaustive superset against coverage.py 7.14.1's matcher. The `source = .` fallback preserves correctness (post-processing still filters).

## Verification
1476 passed / 2 skipped · mypy clean · ruff clean · adversarial QA pass (the bracketed-dir test is mutation-verified to pin the fallback) · python + test purists clean.

### DIFF
--- a/src/pytest_gremlins/plugin.py
+++ b/src/pytest_gremlins/plugin.py
@@ -1781,12 +1781,14 @@ def _run_tests_with_coverage(
             ``dynamic_context=test_function``), the map is used to expand it
             to every matching full node ID so downstream lookups succeed.
         coverage_include: Optional list of file paths to scope coverage
-            tracing to.  When non-empty, the generated coveragerc uses an
-            ``include =`` list of these paths instead of ``source = .``, so
-            coverage.py only traces the gremlin source files rather than the
-            entire tree.  Downstream post-processing discards coverage for
-            non-gremlin files anyway, so narrowing scope here avoids wasted
-            tracing and SQLite I/O without changing the resulting coverage map.
+            tracing to.  When non-empty and contains no glob-special characters
+            (``*?[]``), the generated coveragerc uses an ``include =`` list of
+            these paths instead of ``source = .``, so coverage.py only traces
+            the gremlin source files rather than the entire tree. If any path
+            contains a glob-special character, falls back to ``source = .``
+            (since coverage.py interprets ``include`` entries as fnmatch globs).
+            Downstream post-processing discards non-gremlin coverage, so narrowing
+            scope avoids wasted tracing and SQLite I/O without changing the result.
 
     Returns:
         Dict mapping test names to their coverage data (file path -> lines).
@@ -1800,7 +1802,7 @@ def _run_tests_with_coverage(
     # escaped for coverage.py's tokenizer. When any glob-special char is
     # present, fall back to source = . and rely on post-processing filtering
     # (lines ~1698-1716) to discard non-gremlin coverage.
-    has_glob_special_chars = coverage_include and any(ch in p for p in coverage_include for ch in '*?[]')
+    has_glob_special_chars = any(ch in p for p in (coverage_include or []) for ch in '*?[]')
     if coverage_include and not has_glob_special_chars:
         include_lines = '\n'.join(f'    {path}' for path in coverage_include)
         coveragerc_content = f'[run]\ninclude =\n{include_lines}\n'
--- a/src/pytest_gremlins/plugin.py
+++ b/src/pytest_gremlins/plugin.py
@@ -1680,10 +1680,13 @@ def _collect_coverage(gremlin_session: GremlinSession, rootdir: Path) -> None:
     # Pytest node IDs can be absolute paths in some contexts (e.g., pytester)
     relative_node_ids = _make_node_ids_relative(test_node_ids, rootdir)
 
+    coverage_include = sorted({str(Path(gremlin.file_path).resolve()) for gremlin in gremlin_session.gremlins})
+
     coverage_data = _run_tests_with_coverage(
         relative_node_ids,
         rootdir,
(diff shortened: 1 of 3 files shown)


mikelane (commented):
## PR #423 — cleanup: scope cold-run coverage to gremlin source files

**Verdict: APPROVE**

### Summary

Correct, well-scoped cleanup. The change narrows coverage.py's tracing scope from `source = .` to an explicit `include =` list of resolved gremlin source paths, yielding ~25% smaller coverage DB with no behavioral change to mutation results (post-processing already discards non-gremlin coverage). The glob-special-char fallback is the critical correctness invariant and it is properly proven.

---

### Phase 0 — Bug verification

No bugs found. The adversarial QA finding (bracketed-directory false-survivor, issue #387 class) was correctly diagnosed and fixed: paths containing `*?[]` cannot be safely used as coverage.py `include` entries because they are interpreted as fnmatch globs. The fallback to `source = .` for those paths is the right escape hatch.

The predicate at `plugin.py:1805` is a correct exhaustive superset check — any glob-special character in any path in the list triggers the fallback for the entire include list. Conservative but safe.

### Phase 1 — Quality tools

Lint: SUCCESS, Type Check: SUCCESS, Documentation: SUCCESS, Security Audit: SUCCESS (all from live CI). Tests in progress across 12 matrix configurations (4 Python × 3 OS).

### Phase 2 — Config integrity

No suppressions, no ignore directives added.

### Phase 3 — Code standards

`plugin.py:1807` — the multiline INI continuation format (`include =` followed by 4-space-indented paths) is valid configparser syntax that coverage.py handles correctly.

`coverage_include or None` at the call site in `_collect_coverage` cleanly converts an empty list to `None`, keeping the `_run_tests_with_coverage` contract unambiguous.

### Phase 4 — Self-documenting code

The `# coverage.py's include directive ...` comment at `plugin.py:1800-1804` explains WHY the fallback exists and references the post-processing lines by number. This is exactly the right kind of comment — external constraint, not a restatement of the code.

### Phase 5 — Test quality

**`it_records_coverage_for_source_in_bracketed_directory`** — real integration test. Actually invokes `_run_tests_with_coverage` with a subprocess, exercises the fallback path end-to-end, and asserts on observable behavior (source file appears in covered set). No private method access. Proves the #387-class false-survivor regression is fixed. Excellent.

**`DescribeCollectCoverageScoping.it_passes_resolved_gremlin_source_files_as_coverage_include`** — correctly uses `MagicMock(spec=['file_path'])` (spec enforced per python-conventions.md). Proves the wiring between `_collect_coverage` and `_run_tests_with_coverage` through the public function interface.

**`it_uses_source_dot_when_no_coverage_include`** and **`it_writes_include_section_when_coverage_include_provided`** — contract tests for the coveragerc generation, appropriately unit-scoped using subprocess capture.

### Phase 8 — Regression vs main

Branch divergence point is `bf7bf7c` = current `origin/main` tip. Zero commits have landed on main since divergence. No regression risk.

---

### Praise

The adversarial QA loop produced a genuinely important correctness fix rather than theoretical edge-case hardening. The bracketed-directory test goes RED without the fallback — this is mutation-verified contract coverage, not coverage inflation. The comment explaining the fallback's WHY (fnmatch glob semantics in coverage.py's INI tokenizer, with a line reference to the post-processing that makes the fallback safe) is the ideal documentation style for non-obvious constraints.

---

*Solo-developer repo — formal self-approval not possible; thorough adversarial review comment is the documented substitute per CLAUDE.md.*

Issue #113, opened 2026-02-12
Coverage-guided test selection silently fails when pytest-cov is active


mikelane:
## Summary

When a project has `--cov` flags in pytest's `addopts` configuration (standard practice), the coverage-guided test selection produces an empty coverage map. Every gremlin reports "0 tests cover this gremlin" and the kill rate is 0%.

## Root Cause

`_run_tests_with_coverage` (`plugin.py:672`) runs:

```python
cmd = [
    sys.executable, '-m', 'coverage', 'run',
    f'--rcfile={coveragerc_path}',
    '-m', 'pytest', *test_node_ids, '--tb=no', '-q',
]
```

If the user has `--cov=src/mypackage` in their pytest `addopts`, pytest-cov loads inside the subprocess and takes over coverage measurement. The outer `coverage run` wrapper collects **zero data**. Coverage.py itself emits a warning:

```
CoverageWarning: No data was collected. (no-data-collected)
```

...but this is swallowed because the subprocess runs with `capture_output=True`.

## Why It's Silent

1. `capture_output=True` (line 689) discards stderr, which contains the `CoverageWarning`
2. `except (sqlite3.Error, OSError): pass` (line 734) silently eats database errors
3. An empty coverage map is technically valid — the plugin falls through to "no tests cover this gremlin" for every mutation

The user sees 0% kill rate with no indication of what went wrong.

## Reproduction

```toml
# pyproject.toml
[tool.pytest.ini_options]
addopts = "--strict-markers --cov=src/mypackage --cov-report=term-missing --cov-branch"

[tool.pytest-gremlins]
paths = ["src/mypackage"]
```

```bash
pytest --gremlins
# Every gremlin shows: "0 tests cover this gremlin"
# Kill rate: 0%
```

## Workaround

Move `--cov` flags out of `addopts` and into a task runner command instead:

```toml
[tool.pytest.ini_options]
addopts = "--strict-markers"

[tool.poe.tasks]
test = "pytest --cov=src/mypackage --cov-report=term-missing --cov-branch"
```

## Proposed Fix

Add `'--no-cov'` to the subprocess command in `_run_tests_with_coverage`:

```python
cmd = [
    sys.executable, '-m', 'coverage', 'run',
    f'--rcfile={coveragerc_path}',
    '-m', 'pytest', *test_node_ids, '--tb=no', '-q', '--no-cov',
]
```

[...]

PR #436, opened 2026-06-22
fix(coverage): preserve non-cov addopts in subprocess runs (#424)

### LINKED ISSUE #424: -o addopts= in coverage pre-scan strips --import-mode (and other addopts), silently disabling coverage-guided selection — regression from the #113 fix
## Summary

The fix for #113 added `-o addopts=` to the coverage pre-scan subprocess (`_run_tests_with_coverage`) to stop `pytest-cov` from loading. But clearing addopts **wholesale** also discards any other options projects keep there — notably **`--import-mode=importlib`**. For repos that require a non-default import mode (e.g. colocated test files that share a basename across packages), the pre-scan then fails to collect (`import file mismatch`) → **no coverage data** → gremlins silently falls back to running the entire test set against every mutant. Mutation results stay correct, but the advertised coverage-guided speedup never engages.

Same symptom as #113 (`Coverage collection returned no data`), opposite trigger: there is no `--cov` involved — the loss is the import mode. The warning text still points users at `--cov`, which is misleading here.

## Version

- Reproduced on `pytest-gremlins==1.8.1`, `pytest==9.0.3`, Python 3.14.5.
- **Still present on `main`** (checked 2026-06-08): the pre-scan command still passes `-o addopts=` (≈`plugin.py` L2706/L2717). `--no-cov` is now appended alongside it, but the wholesale addopts clear — and thus the import-mode strip — is unchanged. (Warning at `plugin.py:1649` in 1.8.1.)

## Minimal repro

`pytest.ini`:
```ini
[pytest]
addopts = --import-mode=importlib
pythonpath = .
```
Same-basename test files in two packages **without** `__init__.py` (the case that *requires* importlib mode):
```
pkg_a/calc.py      def add(a, b): return a + b
pkg_a/test_calc.py from pkg_a.calc import add; def test_add(): assert add(2, 3) == 5
pkg_b/calc.py      def mul(a, b): return a * b
pkg_b/test_calc.py from pkg_b.calc import mul; def test_mul(): assert mul(2, 3) == 6
```
```bash
pytest --gremlins --gremlin-targets=pkg_a,pkg_b
```
→ `UserWarning: Coverage collection returned no data ... issue #113`, and every gremlin reports `running 2/2 tests` — i.e. the full suite per mutant; coverage-guided selection never engaged. (No `pytest-cov` installed — `--cov` is not involved.)

## Confirmation it's the import mode

```bash
# The pre-scan's exact action — clearing addopts reverts to prepend mode → same-basename collision:
pytest -o addopts= pkg_a/test_calc.py pkg_b/test_calc.py --co
#   E   import file mismatch: ... pkg_a/test_calc.py is not the same as pkg_b/test_calc.py
#   HINT: ... use a unique basename     -> 1 error, collection interrupted

# The env var survives -o addopts=, restoring the mode:
PYTEST_ADDOPTS="--import-mode=importlib" pytest -o addopts= pkg_a/test_calc.py pkg_b/test_calc.py --co
#   -> 2 tests collected
```

## Workaround (for other affected users)

Run gremlins with `PYTEST_ADDOPTS="--import-mode=importlib"`. The env var is honored independently of the ini `addopts` that `-o addopts=` clears, and the pre-scan subprocess inherits it.

## Suggested fix

Preserve `--import-mode` across the pre-scan rather than relying on the wholesale clear. Keep clearing addopts (so unrelated user addopts can't perturb the pre-scan), but re-inject the configured import mode into the subprocess command — e.g. read it via `config.getoption("importmode")` and append `--import-mode=<mode>` in **both** pre-scan command paths.

*(An earlier draft of this report suggested simply dropping `-o addopts=` — since `main` already appends `--no-cov`, that does neutralize the original #113 `--cov` interference. But dropping the clear wholesale would re-expose the pre-scan to **all** other user addopts — `-n auto` (xdist), `-p no:<plugin>`, `--strict-config`, etc. — which is what the blunt clear was protecting against. Re-injecting only `--import-mode` is the surgical fix that keeps #113 fixed without that risk.)*

disi-omni:
## Summary

The coverage pre-scan (`_run_tests_with_coverage`) and the per-mutant execution runs (`_build_test_command`) override the project's pytest `addopts` with `-o addopts=` so pytest-cov can't hijack coverage collection (#113). Clearing **all** addopts also drops collection-affecting options the project depends on — most notably `--import-mode=importlib` — which silently disengages gremlins' coverage-guided test selection (#424). This strips only the pytest-cov option family and preserves everything else.

## Changes

- Add `_addopts_without_cov(...)`: strips only the `--cov*` / `--no-cov*` option family, computed once in `pytest_configure` from `config.getini("addopts")`, stored on `GremlinSession.preserved_addopts`, and threaded into all three subprocess command builders as `-o addopts=<filtered>`.
- `_is_cov_addopt` matches the exact `--cov` / `--no-cov` switches and their `--cov-*` / `--no-cov-*` families (with or without inline `=value`), so an unrelated option sharing the prefix (e.g. `--covariance-threshold`) is not stripped.
- A cov option written with a space-separated value (`--cov-report term`, `--cov-precision 2`) has its value token dropped too; the value-less switches (`--cov-branch`, `--cov-append`, `--cov-reset`, `--no-cov`, `--no-cov-on-fail`) keep their following token. Unknown cov options default to value-taking so they never orphan a value.
- Default stays `addopts=` (clear-all) when no addopts are configured — behavior unchanged for projects without addopts.
- Known limitations, documented in the docstring (both were also dropped by the old blanket clear, so this is strictly better on the common case): a `--cov` hidden inside a pytest `@argfile` is not introspected, and bare positional addopts are preserved verbatim.

## Type of Change

## Related Issues

Closes #424
Related to #113

## Test plan

New `tests/plugin/test_plugin_addopts_preserve.py` covers `_addopts_without_cov` (inline / space-separated / flag-only / `--cov-precision` / `--cov-reset` / prefix over-match / quoting), the threading into both subprocess command builders, and `pytest_configure` storing the filtered value on the session. The `tests/conftest.py` config double gains a `getini` method (now exercised by `pytest_configure`). Full suite green locally (`tests/plugin tests/config tests/cache tests/coverage_module` + the rest); ruff / mypy / isort clean.

## Testing

- [x] All existing tests pass
- [x] Added new tests for this change
- [x] Manually tested

## Checklist

- [x] Tests written FIRST (TDD)


src/pytest_gremlins/plugin.py
```diff
@@ -1760,12 +1768,86 @@ def _populate_coverage_from_line_bits(
...
+        token = raw_addopts[index]
+        index += 1
+        if _is_cov_addopt(token):
+            # A value-taking cov option written without an inline ``=`` consumes the
+            # next token as its value; drop that too — unless this is a known value-less
+            # switch or the next token is itself an option.
+            if (
+                '=' not in token
+                and token not in _COV_FLAG_ONLY_OPTS
+                and index < count
```
mikelane, line 1835: **MEDIUM — untested final-token boundary**

The `index < count` guard at line 1835 handles inputs like `["--cov"]` (bare `--cov` as the last token) or `["--cov-report"]` (value-taking option with no following token). Those are realistic user configs. A mutant flipping this guard to `index <= count` raises `IndexError` on exactly those inputs and would survive the current suite.

Two lines would cover it:

```python
def it_drops_bare_cov_at_end_of_list() -> None:
    assert _addopts_without_cov(["--import-mode=importlib", "--cov"]) == "--import-mode=importlib"
```

The guard is correct — the test just does not exist yet. Recommend filing as a fast-follow issue.

src/pytest_gremlins/plugin.py
```diff
@@ -1760,12 +1768,86 @@ def _populate_coverage_from_line_bits(
...
+# leaving an orphaned value token behind; the explicit set below stops the known
+# value-less switches from swallowing a following unrelated token.
+_COV_FLAG_ONLY_OPTS = frozenset(
+    {
+        '--cov-branch',
+        '--cov-append',
+        '--cov-reset',
+        '--no-cov',
+        '--no-cov-on-fail',
+    }
```
mikelane, line 1784: **LOW — hardcoded mirror of pytest-cov option set, no drift guard**

`_COV_FLAG_ONLY_OPTS` duplicates knowledge of which pytest-cov options take no value. If pytest-cov adds a new flag-only switch (e.g. `--cov-no-header`), it will be mis-classified as value-taking and silently consume the following token.

The risk is low today — these options are stable — but a comment naming the pytest-cov version this list was sourced from would make future drift detectable at a glance.

src/pytest_gremlins/plugin.py
```diff
@@ -1760,12 +1768,86 @@ def _populate_coverage_from_line_bits(
...
+        index += 1
+        if _is_cov_addopt(token):
+            # A value-taking cov option written without an inline ``=`` consumes the
+            # next token as its value; drop that too — unless this is a known value-less
+            # switch or the next token is itself an option.
+            if (
+                '=' not in token
+                and token not in _COV_FLAG_ONLY_OPTS
+                and index < count
+                and not raw_addopts[index].startswith('-')
```
mikelane, line 1836: **MEDIUM — value-consumption heuristic: next-token-starts-with-dash check**

When a value-taking cov option is written without inline `=` (e.g. `--cov src`), the heuristic at line 1836 drops the next token unless it starts with `-`. This is correct for the common case but creates an untested ambiguity: if the cov value itself starts with `-` (unusual but valid with some shell quoting), the value leaks as an argv token rather than being consumed. This is the same silent-disengage class as issue #424 but for a different edge. A docstring note on the ambiguity and a single parametrized test case would bound the risk.

src/pytest_gremlins/plugin.py
```diff
@@ -745,6 +751,7 @@ def pytest_configure(config: pytest.Config) -> None:
             explain_gremlin_id=getattr(config.option, 'gremlin_explain', None),
             xdist_active=xdist_active,
             xdist_workers=xdist_worker_int if xdist_active else None,
+            preserved_addopts=_addopts_without_cov(config.getini('addopts')),
```
mikelane, line 754: **LOW — `config.getini("addopts")` typed as `Any`, narrowed implicitly**

pytest's `getini` return type is `Any`. The function passes the result directly to `_addopts_without_cov(raw_addopts: list[str])` with no runtime guard. If `addopts` is ever a string (older pytest versions or unusual configurations), the `while index < count` loop iterates over characters, not tokens.

A one-line guard (`if isinstance(raw_addopts, str): raw_addopts = raw_addopts.split()`) or a type assertion would make the contract explicit.

mikelane (commented):
## APPROVED — fix(coverage): preserve non-cov addopts in subprocess runs (#424)

**Verdict: APPROVE**

### Summary

This PR fixes a real silent regression: `_run_tests_with_coverage` and `_build_test_command` were overriding pytest `addopts` with `-o addopts=`, which stripped ALL addopts including `--import-mode=importlib`. That silently disengaged gremlins' coverage-guided selection. The fix is correct, secure, and strictly better than the status quo.

### Quality assessment by phase

**Phase 0 — Bugs:** No bugs found. The bash-quoting specialist independently proved via adversarial round-trip (16 token classes including `$(id)`, backticks, newlines, multibyte UTF-8, empty string) that `shlex.quote` → join → pytest `shlex.split` is lossless and no injection surface exists. The `index < count` guard is correct and necessary.

**Phase 1 — Tooling:** mypy clean, ruff clean, 1495 passed / 0 failed / 2 skipped on rebased code. No CI ran on the fork (external contributor; workflows unapproved) — local run is the only signal.

**Phase 2 — Config integrity:** No exclusions, no linter suppressions, no ignore comments introduced.

**Phase 3 — Code standards:** `shlex.quote` for shell safety, list-form subprocess (no `shell=True`), `frozenset` for O(1) membership, clean type annotations throughout. The `shlex` import is added appropriately. `_addopts_without_cov` is a pure function with no side effects — easily testable and revertable.

**Phase 4 — Self-documenting code:** Docstrings explain *why* (issue #113 / #424 context), not just what. The `_COV_FLAG_ONLY_OPTS` comment block explains the design contract clearly. The known-limitations section in the docstring is honest and appropriate.

**Phase 5 — Test quality:** 14 parametrized-style unit tests for `_addopts_without_cov` all assert on exact output (no `assert True`, no private-method access). The `conftest.py` `_Config.getini` mock raises on unknown ini names, which is an excellent fidelity improvement. The 3 tests for `_build_test_command` and 2 for `_run_tests_with_coverage` cover the threading path. See inline comments for gaps.

**Phase 6 — Architecture:** Pure function + parameter threading pattern. No new abstractions. GremlinSession gains one `str` field with a sensible default. Backward compatibility preserved (default `preserved_addopts=''` maps to the historical clear-all behavior). Blast radius is exactly the intended beneficiaries: projects using addopts + coverage selection.

**Phase 7 — Security:** CLEAN. No injection surface. `shlex.quote` handles all adversarial input classes. No hardcoded secrets. No shell=True.

**Phase 8 — Regression vs main:** PR was rebased onto current main. Conflict-free. Changed files: `plugin.py`, `tests/conftest.py`, `tests/plugin/test_pardoned_gremlin_execution.py` (arity fix), `tests/plugin/test_plugin_addopts_preserve.py` (new). No overlap with main's recent commits (uv.lock / workflows only). No regression risk.

### Findings (all non-blocking)

| Severity | Location | Finding |
|---|---|---|
| MEDIUM | plugin.py:1835 | Final-token boundary untested — `["--cov"]` as last token exercises the `index < count` guard but no test covers it |
| MEDIUM | plugin.py:1836 | Value-heuristic: next-token-starts-with-dash check has an untested ambiguity for cov values beginning with `-` |
| MEDIUM | plugin.py:754 | `config.getini("addopts")` typed `Any`; no runtime guard before passing to `list[str]`-typed function |
| LOW | plugin.py:1784 | `_COV_FLAG_ONLY_OPTS` is a hardcoded mirror with no drift guard vs. pytest-cov's actual option set |
| LOW | plugin.py:1836 | Value-consumption heuristic is undocumented in the docstring as a known ambiguity |
| LOW | (general) | No end-to-end committed regression test proving pytest honors the re-injected `--import-mode=importlib` (behavior verified live by adversarial specialist; a committed test would lock it in) |

### Praise

The `_addopts_without_cov` design is clean: pure function, composed with `shlex.quote`, no subprocess knowledge baked in. The PR description's "strictly better than the old blanket clear" framing is accurate and appropriately scoped. The `conftest.py` improvement (raising on unknown ini names rather than returning falsy defaults) is a quality improvement beyond the immediate fix. The arity fix in `test_pardoned_gremlin_execution.py` (`lambda _: ['pytest']` → `lambda *_: ['pytest']`) is exactly the right minimal fix and catches a latent test brittleness.

### Merge prerequisites

**This PR is behind main.** The fork branch `cf0462ef` has not been rebased for CI. Before merge:

1. Rebase onto current main (conflict-free — verified locally on the rebased worktree)
2. Push the rebased branch — the new SHA will invalidate this verdict; a re-review at the rebased SHA will be needed before the merge hook allows merge
3. Recommend filing a follow-up issue for the 3 MEDIUM findings (untested boundary cases); ~4 lines of test total

No CI ever ran on this PR (external fork, workflows unapproved). The local run (1495 passed) is the only signal. Consider approving the workflow run for the fork after rebase.

Issue #442, opened 2026-06-29
ci: pip-audit fails on main — bump msgpack→1.2.1 and pip→26.1.2 (latent CVEs blocking all PRs)


mikelane:
## Problem

The `Security Audit` CI job (`uv run pip-audit`) is failing on **every open PR** — including all 9
dependabot PRs (#425, #430, #431, #434, #437, #438, #439, #440, #441) and #436. The failures are
**not** caused by the PRs' own changes; they are latent `main`-level failures from two
newly-disclosed transitive CVEs that `pip-audit` flags against the live advisory DB.

```
Found 2 known vulnerabilities, ignored 1 in 2 packages
Name    Version ID                  Fix Versions
msgpack 1.1.2   GHSA-6v7p-g79w-8964 1.2.1
pip     26.1.1  PYSEC-2026-196      26.1.2
```

Both are transitive dependencies (msgpack via a transitive chain; pip via pip-audit). This is the
same failure shape that #418 fixed in April — a CVE disclosed after the lockfile was pinned
retroactively fails all open PRs' CI.

## Fix

Bump the two flagged packages in `uv.lock` only (no source changes):

- `msgpack` 1.1.2 → 1.2.1
- `pip` 26.1.1 → 26.1.2

```bash
uv lock --upgrade-package msgpack --upgrade-package pip
```

## Acceptance

- `uv run pip-audit --ignore-vuln CVE-2026-4539` reports 0 vulnerabilities (1 ignored: pygments
  CVE-2026-4539, no fix available)
- Full test suite still green; mypy + ruff clean
- `uv.lock` is the only changed file

## Impact

Unblocks the entire dependabot backlog and any future PR — all are currently red on this single
`main`-level Security Audit failure.

PR #418, opened 2026-05-30
chore(deps): patch 13 CVEs in transitive deps + harden flaky cache-perf test

mikelane:
Closes #417.

Unblocks the dependabot backlog (#390, #399, #405, #406, #408, #409, #411, #412, #413), all of which were failing CI on these two latent `main` problems — not on their own bumps.

## Changes

**Security (uv.lock only):** bumped 6 transitive deps to fix versions, patching 13 newly-disclosed 2026 CVEs that `pip-audit` flags against the current advisory DB (idna→3.17, pillow→12.2.0, pip→26.1.1, pymdown-extensions→10.21.3, urllib3→2.7.0, uv→0.11.17; plus pytest→9.0.3 for CVE-2025-71176). `pip-audit` now reports 0 vulnerabilities (1 ignored: pygments CVE-2026-4539, no fix available).

**Flaky test:** `it_skips_test_execution_on_cache_hit` no longer asserts raw wall-clock (`warm_time < cold_time`, which lost by ~16ms under CI variance); it asserts the cache-hit *behavior* (`*cache hit*` in output, proving execution was skipped). 5/5 deterministic passes locally.

## Verification
mypy clean · ruff clean · 1450 passed / 2 skipped · pip-audit 0 vulns

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #447, opened 2026-06-30
docs: add v1.9.0 changelog entry

### LINKED ISSUE #424: -o addopts= in coverage pre-scan strips --import-mode (and other addopts), silently disabling coverage-guided selection — regression from the #113 fix

mikelane:
## What

Adds the `v1.9.0` CHANGELOG entry ahead of cutting the release, plus two cspell dictionary
additions (`GHSA`, `PYSEC`) so security-advisory references pass spell-check.

## Why

Per the project release process, the CHANGELOG must be updated and merged to main **before**
dispatching `cut-release.yml` — the workflow's releasable-commits guard is satisfied by this commit.

## Version: v1.9.0 (minor)

A `feat:` commit landed since v1.8.1 (`--gremlin-explain`, #398), so the bump resolves to a
**minor** (1.9.0), not a patch. The entry covers:
- **Feat**: `--gremlin-explain` diagnostic flag (#398)
- **Fix**: non-cov addopts preservation (#436, closes #424); parametrize-pipe split (#419)
- **Internal**: cold-run coverage scoping (#423); security CVE bumps (#418, #442); dependency wave

## Scope

Docs-only — `CHANGELOG.md` + `cspell.json`. No executable code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,36 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.9.0 (2026-06-29)
+
+### Feat
+
+- **`--gremlin-explain <id>`**: new diagnostic flag that prints why a specific gremlin was
+  selected, instrumented, and which test was chosen to run against it — exposing how
+  coverage-guided selection reaches its decisions (#397, #398)
+
+### Fix
+
+- **Preserve non-cov `addopts` in subprocess runs**: gremlins' coverage pre-scan and
+  mutant-execution runs previously cleared *all* pytest `addopts` via `-o addopts=` to neutralize
+  pytest-cov, silently dropping collection-affecting options such as `--import-mode=importlib`.
+  This disabled coverage-guided test selection for projects relying on those options. Only the
+  `--cov*` / `--no-cov*` family is now stripped; everything else survives into the subprocess
+  (#424, #436)
+- **Parametrize values containing `|`**: coverage selection now splits on the last pipe so
+  parametrized test IDs whose values contain `|` are no longer dropped (#419)
+
+### Internal
+
+- **Cold-run coverage** scoped to gremlin source files, avoiding wasted tracing on unrelated
+  modules (#423)
+- **Security**: bumped transitive dependencies to clear latent CVEs that `pip-audit` flagged
+  against the live advisory DB — 13 CVEs in #418 (idna, pillow, pip, pymdown-extensions, urllib3,
+  uv, pytest) and msgpack→1.2.1 + pip→26.1.2 in #442/#443 (GHSA-6v7p-g79w-8964, PYSEC-2026-196)
+- **Dependencies**: routine bumps across the dev toolchain — pytest 9.1.1, ruff 0.15.20,
+  mypy 2.1.0, tox 4.56.1, coverage 7.14.3, commitizen 4.16.4, pip-audit 2.10.1, playwright 1.61.0,
+  codecov-action 7, plus several GitHub Actions group updates
+
 ## v1.8.1 (2026-04-02)
 
 ### Fix
--- a/cspell.json
+++ b/cspell.json
@@ -14,6 +14,8 @@
   ],
   "words": [
     "ASGI",
+    "GHSA",
+    "PYSEC",
     "Dsonar",
     "Qube",
     "sonarqube",


mikelane (commented):
APPROVE (docs-only scope)

## Review Summary

Scope confirmed: only `CHANGELOG.md` and `cspell.json` changed. No executable code touched.

### Changelog accuracy — verified against commit log since v1.8.1

| Claim | Evidence |
|---|---|
| `--gremlin-explain` feat (#397, #398) | `bf7bf7c feat: add --gremlin-explain diagnostic flag (#397) (#398)` — correct; drives the minor bump |
| Non-cov addopts preservation (#424, #436) | `ae29011 fix(coverage): preserve non-cov addopts in subprocess runs (#424) (#436)` — correct |
| Parametrize pipe fix (#419) | `daf9678 fix: rsplit on last pipe so parametrize values containing | are not dropped from coverage selection (#419)` — correct |
| Cold-run coverage scoping (#423) | `058990f cleanup: scope cold-run coverage to gremlin source files (#423)` — correct |
| Security CVE bumps (#418, #442/#443) | `a2e6705` and `a9dc77c` — correct; GHSA/PYSEC identifiers verified |
| Dep wave (pytest 9.1.1, ruff 0.15.20, mypy 2.1.0, tox 4.56.1, etc.) | Multiple `chore(deps):` commits confirm all versions listed |

**Version bump rationale**: `feat:` commit (`--gremlin-explain`) landed since v1.8.1, so minor bump to v1.9.0 is correct per semver.

### cspell.json additions

`GHSA` and `PYSEC` added as dictionary words — necessary so the advisory IDs in the changelog pass spell-check. Legitimate and minimal.

### Format compliance

Follows the existing Keep-a-Changelog `## vX.Y.Z (date)` + `### Feat/Fix/Internal` convention. Consistent with surrounding entries.

### CI

All checks passed: Spell Check, Lint Markdown, Build & Validate, Documentation, All Checks Passed. Code-only checks (Lint, Type Check, Test, Security Audit) correctly skipped for a docs-only change.

### Findings

None. 0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW.

---

Ready to merge. Dispatch `cut-release.yml` after this lands to trigger the v1.9.0 release.

PR #448, opened 2026-07-03
Update comparison.md

boxed:
The comparison page was pretty bad. It sounds like it's AI slop honestly. This fixes the absolute worst parts of the comparisons to mutmut that was misleading, wrong, irrelevant or just word salad. (I'm the original author of mutmut btw)


mikelane:
Thanks for taking the time to sharpen this page, Anders — and for doing it in good faith on a comparison that features your own tool. A few of these are clear improvements I'm happy to take:

- Removing the duplicated "Unix/WSL required" bullet (it was already covered by the "Unix only" line below it).
- "`mutmut apply` fixes surviving mutants" → "try out" — you're right, that wording overstated what `apply` does.
- The "and for state isolation" clarification on the `fork()` description.

Two changes I'd like to talk through before merging rather than take as-is:

1. **"Mutation switching" → "Mutation schemata."** I know *mutation schemata* (Untch et al., 1993) is the precise academic term for the shared technique, and it's fair to note both tools are in that family. But "Mutation switching" is the language we use throughout our own docs and README to describe our specific implementation, so I'd rather keep our term in the pytest-gremlins column (and I'm open to a footnote or intro sentence naming the schemata lineage for both tools, if you think that's clearer for readers).

2. **Deleting the workflow diagram.** The side-by-side "mutmut workflow / pytest-gremlins workflow" ASCII block is one of the more useful bits for a reader trying to understand the mechanical difference at a glance. If any step in the mutmut column is inaccurate, I'd much rather fix that step than drop the whole diagram — corrections there are very welcome.

I'll leave this open so we can settle those two points. Everything else looks good, and I appreciate the accuracy pass. 🙏

boxed:
1. I'd argue that you should change the rest of your docs then. "Mutation switching" could be both mutation schemata or on disk mutation one-by-one like mutmut 1.0 did. Standard nomenclature is good for a reason. 

2. The "workflow" list didn't make sense. I tried fixing it so that it was factually correct first. The issue is that the list was a) incorrect, and worse b) bot a description of how either the system works when working through mutants or how the users workflow for using the tool should be.

mikelane:
You've convinced me on both counts, Anders. Taking this as-is.

**On the terminology** — you're right, and my objection was weaker than I made it sound. Our own architecture doc describes the technique as "instrument the code once with all mutations embedded, then use an environment variable to select which mutation is active," which is Untch et al. 1993 in all but name. And your point that "mutation switching" is ambiguous between schemata and mutmut 1.0's on-disk approach is exactly the kind of ambiguity a comparison page shouldn't be introducing. Having both columns read "Mutation schemata" and putting the real difference where it actually lives — worker process vs `fork()` — is a better table than the one I was defending. That framing also makes our cross-platform story *more* legible, not less, so I was arguing against my own interests.

I'll open a separate issue about whether to align the rest of the docs on the standard term. That's a bigger change and a separate decision from this page.

**On the diagram** — fair. I misread your intent as "this step is wrong" when your actual objection was that the framing was broken: it was neither a description of how either tool works internally nor of how a user works with the tool, and you'd already tried to salvage it before concluding it wasn't salvageable. That's a reasonable call and you're better placed than I am to make it about the mutmut column. Better to have no diagram than a confidently-wrong one on a page that advertises itself as a fair comparison. If I add an internals diagram back later, it'll be clearly scoped as pytest-gremlins internals rather than a side-by-side.

Thanks for the accuracy pass, and for doing it on a page that compares your tool to mine. That's a good-faith contribution and I appreciate it. 🙏

mikelane (commented):
## Review: docs-only comparison.md update (external fork, boxed/mutmut)

Reviewed at head SHA `e1f5ed264df4d78c0c24281f109993bbbe3d206e`.

### Scope verification

- Confirmed via `gh api .../pulls/448/files` and diff inspection: **exactly one file changed** — `docs/guide/comparison.md` (+3/-20). No executable code, no `.github/workflows/*`, no `pyproject.toml`, no lockfile, no scripts. No supply-chain or workflow-injection vector from this external fork.
- Markdown well-formed: table column counts match the header row for the edited "Speed Architecture" row, all fenced code blocks in the file remain balanced (the deleted block was a matched pair, cleanly removed), and no orphaned references to the deleted "mutmut workflow / pytest-gremlins workflow" ASCII diagram exist anywhere else in `docs/` (grepped the tree — only the unrelated cosmic-ray workflow diagram further down the same file remains, untouched).
- Surrounding prose reads coherently after the diagram deletion — the paragraph before ("The key architectural difference is in isolation...") flows directly into "The practical speed difference depends on your project..." with no dangling "as shown above" references.
- No new factually false claims about pytest-gremlins introduced. The maintainer's PR-comment thread with the author (an informed, deliberate back-and-forth, not a rubber stamp) already settled the two substantive points: the "Mutation schemata" terminology change and the diagram deletion.

### One finding — LOW / informational (already tracked, not a blocker)

The top "Quick Comparison" table now calls pytest-gremlins' technique "Mutation schemata (worker process based)" (line 12), but the "Speed Optimizations" table further down the *same file* (line 142) and the prose immediately below it (line 148: "**Mutation switching** is the key architectural difference...") still say "Mutation switching." This is a real internal terminology split within one document — but per the PR discussion, the maintainer explicitly recognized this and committed to a separate follow-up issue to align terminology across the rest of the docs, calling it "a bigger change and a separate decision from this page." Recording this as LOW/deferred rather than blocking on it, since re-litigating an already-made editorial call isn't this review's job. No such follow-up issue exists yet as of this review — worth opening one to track it.

### Important operational finding — not a content defect, surfacing for the merge gate

**CI has not actually completed on the current head SHA.** `gh api repos/mikelane/pytest-gremlins/commits/e1f5ed2.../check-runs` and `gh run list` show both the `CI` and `Documentation` workflow runs on `e1f5ed264df4d78c0c24281f109993bbbe3d206e` sitting in `conclusion: action_required` — the standard GitHub gate requiring a maintainer to approve workflow runs for an external, first-time contributor after `gh pr update-branch` reset the head SHA. `mergeStateStatus` is `BLOCKED`, not `CLEAN`. Branch protection on `main` requires `Lint`, `Type Check`, `Documentation`, `Security Audit`, and `All Checks Passed` — none of those have actually run on this SHA yet.

This is **not** a CI failure — the same content passed cleanly pre-rebase on commit `5167c5c` (`CI`: success, `Documentation`: success). It just needs a maintainer to approve the pending workflow runs on the current SHA before merge-guardian can verify green and proceed. Flagging this because the review brief assumed CI was already green, and mechanical verification showed otherwise — didn't want that gap to reach the merge gate silently.

### Verdict

**APPROVE** (content). 0 CRITICAL, 0 HIGH, 0 MEDIUM, 1 LOW (informational, deferred by maintainer). No blockers on the content itself. Merge readiness is gated separately on getting CI to actually run and go green on `e1f5ed2` — that's an operational step for a maintainer/merge-guardian, not a content issue.

Nice, good-faith contribution from the mutmut author on a page that compares against his own tool — the accuracy pass genuinely improved the page, and the maintainer's discussion thread is a good model of settling disagreements with evidence rather than by fiat.

Issue #455, opened 2026-07-11
[Bug]: ContentHasher.hash_file crashes on PEP 263 non-utf-8 source (cache/hasher.py:53)


mikelane:
## Summary

`ContentHasher.hash_file` (`src/pytest_gremlins/cache/hasher.py:53`) reads source with `path.read_text(encoding='utf-8')`. This crashes with `UnicodeDecodeError` on a legitimately non-utf-8 source file — e.g. a Python module with a `# -*- coding: latin-1 -*-` declaration (valid per PEP 263) containing bytes that aren't valid utf-8.

This is the same bug **family** as #382 (the `_add_source_file` UnicodeDecodeError), but in a different module — the incremental analysis cache. It was found and reproduced during the adversarial-QA gauntlet for #382 but is out of scope for that fix (which targets `plugin.py`).

## Reproduction

Hashing a latin-1-encoded `.py` source file (with a PEP 263 coding declaration and a byte like `0x81` that's invalid in utf-8) raises `UnicodeDecodeError` from `hash_file`.

## Why the fix is different from #382

#382's fix uses `tokenize.open()` because `_add_source_file` needs to *parse* the source (it must honor the source encoding to get correct text for `ast.parse`).

A **content hasher does not need decoded text** — it needs stable, deterministic bytes. The correct fix here is `path.read_bytes()` (hash the raw bytes directly), which:
- Never raises `UnicodeDecodeError` regardless of encoding
- Is faster (no decode step)
- Produces a stable hash keyed on file content, which is what the incremental cache actually wants

Verify the hash is still stable/consistent with how the rest of the cache keys content after the change.

## Severity

MEDIUM — reachable whenever a project with non-utf-8 source uses the incremental cache. Causes a crash, not silent corruption. Found via #382 gauntlet.

PR #456, opened 2026-07-11
fix(plugin): decode source files via tokenize.open (closes #382)

### LINKED ISSUE #382: [Bug]: UnicodeDecodeError in _add_source_file
### Bug Description

I get the following crash on my code base:

```
INTERNALERROR>   File "C:\Users\Aran-Fey\AppData\Local\Programs\Python\Python314\Lib\site-packages\pluggy\_callers.py", line 121, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "C:\Users\Aran-Fey\AppData\Local\Programs\Python\Python314\Lib\site-packages\pytest_gremlins\plugin.py", line 891, in pytest_collection_finish
INTERNALERROR>     source_files = _discover_source_files(session, gremlin_session)
INTERNALERROR>   File "C:\Users\Aran-Fey\AppData\Local\Programs\Python\Python314\Lib\site-packages\pytest_gremlins\plugin.py", line 975, in _discover_source_files
INTERNALERROR>     _add_source_file(py_file, source_files)
INTERNALERROR>     ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "C:\Users\Aran-Fey\AppData\Local\Programs\Python\Python314\Lib\site-packages\pytest_gremlins\plugin.py", line 1101, in _add_source_file
INTERNALERROR>     source = path.read_text()
INTERNALERROR>   File "C:\Users\Aran-Fey\AppData\Local\Programs\Python\Python314\Lib\pathlib\__init__.py", line 788, in read_text
INTERNALERROR>     return f.read()
INTERNALERROR>            ~~~~~~^^
INTERNALERROR>   File "C:\Users\Aran-Fey\AppData\Local\Programs\Python\Python314\Lib\encodings\cp1252.py", line 23, in decode
INTERNALERROR>     return codecs.charmap_decode(input,self.errors,decoding_table)[0]
INTERNALERROR>            ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR> UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 4999: character maps to <undefined>
```

AFAIK python 3 assumes utf-8 encoding per default, but `pytest-gremlins` does not:

```python
source = path.read_text()
```

So adding `encoding='utf-8'` there should fix it.

### Steps to Reproduce

1. Be on a system where utf-8 is NOT the default file system encoding
2. Add some utf-8 source files to your code base
3. Run `pytest --gremlins`

### Expected Behavior

The code should be decoded/parsed correctly.

### Actual Behavior

The `pytest` command crashes.

### Minimal Reproducible Example

```python

```

### pytest-gremlins Version

1.8.1

### Python Version

3.14.2

### pytest Version

9.0.2

### Operating System

Windows

### Additional Context

Boy, I sure had to fill in a lot of irrelevant information there.

mikelane:
## Summary

Fixes #382 — `UnicodeDecodeError` crash (INTERNALERROR) when pytest-gremlins reads a source file on a non-utf-8-default platform (e.g. Windows cp1252).

## Root cause

`_add_source_file` read source with a bare `path.read_text()`, using the platform default codec. On Windows, a utf-8 source file with non-ASCII characters crashed. The `except` clause caught only `SyntaxError`/`OSError`, so the `UnicodeDecodeError` propagated as an uncaught INTERNALERROR.

## Fix

Reads source via **`tokenize.open()`** — the same mechanism CPython uses to read source files. This:

- Honors **PEP 263 coding declarations** (`# -*- coding: latin-1 -*-`), so legitimately non-utf-8 source files decode with their declared codec rather than crashing.
- Strips a **utf-8 BOM** (utf-8-sig), which previously caused such files to be silently excluded from mutation testing (the BOM made `ast.parse` raise `SyntaxError`).
- Defaults to utf-8 when no declaration is present — fixing the originally reported Windows crash.

The skip clause is broadened to `(OSError, UnicodeDecodeError, LookupError)` so any genuinely undecodable or bad-declaration file is skipped gracefully instead of crashing the session.

## Why not just `encoding='utf-8'`

A hardcoded `encoding='utf-8'` fixes the reported crash but **introduces a regression**: it would crash on a valid latin-1 source file (PEP 263) that the platform-default read previously handled. `tokenize.open()` is strictly more correct — it's what the interpreter itself uses.

## Testing

- Behavioral tests in `tests/plugin/test_plugin_helpers.py` (`DescribeAddSourceFile`): utf-8 with non-ASCII chars, latin-1 with PEP 263 declaration, utf-8-sig BOM.
- Replaced a mechanism-coupled mock test (asserted `read_text` was called with a specific arg) with a behavioral test that asserts the *outcome* (source decodes correctly), so it survives the implementation change.
- `uv run pytest tests -m small` green; mypy clean; ruff clean.

## Found during gauntlet, tracked separately

Adversarial QA found a **sibling** of this bug at `cache/hasher.py:53` (same `read_text(encoding='utf-8')` pattern in the incremental cache). It's out of scope here and needs a different fix (`read_bytes()` for a content hasher) — filed as #455.

Closes #382


src/pytest_gremlins/plugin.py
```diff
@@ -1124,12 +1125,14 @@ def _add_source_file(path: Path, source_files: dict[str, str]) -> None:
         source_files: Dictionary to add the file to.
     """
     try:
-        source = path.read_text()
+        # tokenize.open honors PEP 263 coding declarations and strips BOM
+        with tokenize.open(str(path)) as source_stream:
```
mikelane, line 1129: Nit (LOW): `tokenize.open` already accepts a `Path`-like object (it forwards to `io.open` after encoding detection), so `str(path)` here is a redundant cast — not a defect, just a style nit for a future pass. Not blocking.

mikelane (commented):
## Review: fix(plugin): decode source files via tokenize.open (closes #382)

**Verdict: APPROVE** (posted as COMMENT — solo-authored PR, GitHub disallows self-approval; see local verdict artifact for the authoritative APPROVE record)

### Correctness

The core fix replaces `path.read_text()` with `tokenize.open(str(path))` in `_add_source_file` (plugin.py:1120). This is the right call — `tokenize.open` implements PEP 263 encoding-cookie detection (reads the first two lines for a `# -*- coding: ... -*-` declaration, falling back to utf-8 per PEP 3120) and transparently strips a leading utf-8 BOM. It's used correctly: opened as a context manager, `.read()` called once, closed automatically via `with`. This is exactly the mechanism CPython's own tokenizer/compiler uses to decide source encoding, so behavior now matches what `python foo.py` would do when parsing the same file.

Broadening `except OSError` to `except (OSError, UnicodeDecodeError, LookupError)` is correct and necessary: `tokenize.open` raises `SyntaxError` for a malformed/unknown coding cookie in some cases but raises `LookupError` for an unrecognized codec name and `UnicodeDecodeError` when the declared codec still can't decode the bytes. Without broadening this, an unusual-but-real source file would crash the whole pytest-gremlins session with `INTERNALERROR` instead of gracefully skipping one file — precisely the failure mode described in issue #382 for the Windows cp1252-default case.

The two `open(sources_file)` → `open(sources_file, encoding='utf-8')` changes for the plugin-written JSON files are good symmetrical hardening — the plugin controls the encoding on write (implicitly utf-8 via `json.dump` defaults) so pinning `utf-8` on read removes platform-locale dependence for these internal round-trips. Confirmed the same fix was needed and applied in both `_get_bootstrap_script` and `_get_lightweight_runner_script` generated sources (plugin.py:1321, 1392) — nice catch extending the same class of bug to the generated subprocess scripts, and there's a dedicated `DescribeGeneratedScriptOpenCalls` test class using AST inspection to guard against regression there.

### Test quality

All three new tests in `DescribeAddSourceFile` are behavioral, not implementation-probing:
- `it_reads_utf8_source_with_non_ascii_characters` — writes real utf-8 non-ASCII source, asserts content survives round-trip.
- `it_reads_latin1_source_with_pep263_coding_declaration` — writes real latin-1-encoded bytes with a PEP 263 declaration, asserts correct decode.
- `it_reads_utf8_source_with_bom` — writes real BOM-prefixed bytes, asserts the file is accepted (BOM would otherwise break `ast.parse`).

Manually verified the latin-1 case would fail without the fix: `Path.read_text()` on the same latin-1 bytes throws `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9...` — confirming the test is meaningfully RED against the old implementation, not just green-by-construction. All three tests go through the public `_add_source_file` entry point and assert on its externally observable output (`source_files` dict), not on `tokenize.open` call internals.

`_open_calls_missing_utf8_encoding` / `_is_binary_mode_open` are reasonable small AST-walking helpers scoped to the test file — appropriately excluding binary-mode `open()` calls (which can't take `encoding=`) from the utf-8 requirement.

### Scope

Diff is tightly scoped: one import added, one function body changed, two `open()` calls annotated, and a corresponding test file addition. No unrelated refactors, no drive-by renames, no scope creep.

### CI / regression check

- Merge-base of this branch and `origin/main` is `0802482`, which **is** `origin/main`'s current HEAD — the branch is fully rebased on latest main, so the squash-merge staleness hazard (long-lived branch silently reverting sibling-PR work) does not apply here.
- Lint, type-check (mypy strict), Security Audit, Documentation, and Test on Python 3.11–3.14 for ubuntu-latest and macos-latest are all green. The Python 3.11–3.14 windows-latest matrix jobs were still `IN_PROGRESS` at review time (not failed) — this is the platform the original bug (#382) targets, so it's worth confirming those finish green before merge, but nothing currently indicates failure.
- Locally reran `ruff check` (clean) and `mypy src/pytest_gremlins/plugin.py` (clean, strict) against the worktree head.

### Minor (LOW, non-blocking)

Left an inline nit: `tokenize.open(str(path))` — `tokenize.open` already accepts a path-like object directly, so the `str()` cast is redundant. Purely cosmetic, doesn't affect correctness.

### Praise

The PEP 263 + BOM combination is a genuinely easy trap to fall into (a naive utf-8-only fix, which is literally what commit `7caaa77` on this same branch's history originally shipped, silently regresses on latin-1-declared or BOM-prefixed source). Good catch by the adversarial-QA pass, and the fix reaches for the standard-library tool that already encodes the correct PEP 263 semantics rather than hand-rolling BOM/encoding-cookie detection. The docstrings on the new tests explain *why* each case matters (referencing the specific PEP and the specific crash mode) rather than restating what the code does — good self-documenting test design.

PR #458, opened 2026-07-11
feat(reporting): selected-test column + sonarqube error diagnostics

### LINKED ISSUE #337: fix: add gremlin error diagnostics to all report formats
## Problem

When a gremlin produces a subprocess exit code >= 2 (pytest collection/import/internal error), the error is recorded as "ERROR" status but the subprocess stderr is captured and silently discarded. There is no way for a user to diagnose why gremlins are erroring without reading the pytest-gremlins source code and manually reproducing the subprocess call.

This was discovered running gremlins against attrs v25.3.0, where 680/681 gremlins errored with zero diagnostic output. But the problem affects any project where gremlins encounter errors — the user sees "Error: 680 gremlins (100%)" and has no path to understanding why.

**Why now**: The autoresearch cold-cache optimization work (#336) needs attrs as a benchmark target. Debugging the 100% error rate requires error output to be visible. More broadly, any user hitting gremlin errors today is flying blind.

## Steps to Reproduce

1. Clone attrs: `git clone --depth 1 --branch 25.3.0 https://github.com/python-attrs/attrs.git /tmp/attrs-test`
2. Create venv: `uv venv /tmp/attrs-test/.venv && source /tmp/attrs-test/.venv/bin/activate`
3. Install deps: `uv pip install "attrs==25.3.0" pytest hypothesis -e /path/to/pytest-gremlins`
4. Run gremlins: `cd /tmp/attrs-test && pytest --gremlins --gremlin-targets src/attr --gremlin-workers=1 -q`
5. Observe: "Error: 680 gremlins (100%)" with no stderr, no traceback, no diagnostic info

**Expected**: Error details visible in console output, HTML report, and JSON report.
**Actual**: Silent "Error" count with no actionable information.

## Acceptance Criteria

```gherkin
Scenario: GremlinResult stores subprocess stderr on error
  Given a gremlin subprocess exits with code >= 2
  And the subprocess wrote content to stderr
  When the GremlinResult is created
  Then error_output contains the stderr content (up to 2000 chars, decoded with errors='replace')
  And the error_output field defaults to empty string for non-error gremlins

Scenario: WorkerResult stores subprocess stderr on error (parallel path)
  Given a parallel worker's gremlin subprocess exits with code >= 2
  And the subprocess wrote content to stderr
  When the WorkerResult is created
  Then error_output contains the stderr content (up to 2000 chars)
  And the error_output propagates to the final GremlinResult when results are aggregated

Scenario: Error subprocess output is logged at DEBUG level
  Given a gremlin subprocess exits with code >= 2
  When the result is recorded
  Then a DEBUG log message is emitted with the gremlin ID, exit code, and first 200 chars of stderr

Scenario: Import hook instrumentation errors are logged
  Given the GremlinLoader compiles and runs instrumented AST for a module
  When the instrumented code raises an exception during module loading
  Then an ERROR log message is emitted with the module name and full traceback (exc_info=True)
  And the original exception is re-raised (not swallowed)

Scenario: Console report shows error details with --gremlin-error-detail
  Given gremlins ran with --gremlin-error-detail=N where N > 0
  And at least one gremlin has ERROR status with non-empty error_output
  When the console report is printed
  Then after the "Error: X gremlins (Y%)" summary line
  A "Top error details:" section shows up to N errored gremlins
  Each showing file:line and the last line of stderr (truncated to 120 chars)

Scenario: Console report omits error details by default
  Given gremlins ran without --gremlin-error-detail (default 0)
  When the console report is printed
  Then no "Top error details:" section appears
  And the "Error: X gremlins (Y%)" summary line still appears if there are errors

Scenario: HTML report shows full traceback per errored gremlin
  Given gremlins ran with --gremlin-report=html
  And some gremlins have ERROR status with non-empty error_output
  When the HTML report is viewed
  Then each errored gremlin's detail section includes the full stderr
  And the stderr is rendered in a scrollable, monospaced code block
  And the stderr content is HTML-escaped to prevent XSS

Scenario: JSON report includes error_output field
  Given gremlins ran with --gremlin-report=json
  And some gremlins have ERROR status with non-empty error_output
  When the JSON report is parsed
  Then each gremlin result object includes an "error_output" string field
  And error_output is empty string for non-error gremlins

Scenario: MutationScore provides top_errors() accessor
  Given a MutationScore is created from results that include errored gremlins with error_output
  When top_errors(limit=5) is called
  Then it returns up to 5 GremlinResult objects with ERROR status and non-empty error_output
  And results without error_output are excluded

Scenario: Cached results handle missing error_output gracefully
  Given incremental cache contains results from a prior run without error_output field
  When the cache is loaded in a version that has error_output
  Then the missing field defaults to empty string
  And no deserialization errors occur
```

## Out of Scope

- **Fixing the attrs compatibility bug itself** — that's a separate issue, blocked by this one
- **Adding attrs as a CI integration test** — separate issue, blocked by the attrs fix
- **Changing how errors are classified** (exit code 2 vs 3 vs signal) — current classification is correct, just invisible
- **Storing error_output in the incremental cache** — error_output is diagnostic, not a cached result. It should be available during the run but does not need to persist across runs. The cache stores status (ERROR) which is sufficient for incremental skip logic.
- **Changing the mutation switching architecture** — logging only, no behavioral changes to instrumentation

## Edge Cases

- **Non-UTF-8 stderr**: Use `decode(errors='replace')` — replaces invalid bytes with U+FFFD rather than crashing
- **Empty stderr on error**: Some exit codes produce no stderr (e.g., SIGKILL). `error_output` stays empty string. The console report skips gremlins with empty error_output in "Top error details"
- **Very long stderr**: Truncate to 2000 chars at capture time. HTML report shows full 2000 chars; console shows last line truncated to 120 chars
- **HTML report XSS**: error_output content must be HTML-escaped before rendering (could contain `<script>` in theory)
- **Cache backwards compatibility**: Old cached GremlinResult objects won't have `error_output`. The field defaults to `''`, so deserialization is backwards-compatible

## Dependencies

- Blocks: attrs compatibility fix (not yet filed)
- Blocks: attrs CI integration (not yet filed)
- Blocks: autoresearch cold-cache optimization (#336) — indirectly, via attrs fix

### LINKED ISSUE #414: Show selected test (and outcome) per mutant in the HTML report
When a mutant survives, the report shows the mutation diff but not which test(s) the coverage-based selector ran or what their outcomes were. That makes it hard to distinguish three cases:

- "We ran test_foo and it passed" → genuine survivor, test rigor gap.
- "No test covered this line" → coverage attribution gap.
- "Selector picked a test that doesn't actually exercise the mutated branch, but other tests in the suite would have killed it" → selector quirk, not a real survivor.

I hit the third case this week with return value → None mutants on Pydantic BaseModel factory methods. Without per-mutant test history in the report, all three look identical.

Proposal: in the survivor section of coverage/gremlins/index.html, expand each mutant to list every test the selector chose, the order they ran, and each outcome (pass / fail / errored / not-run). For survivors, this immediately disambiguates the three failure modes.

mikelane:
## Summary

Two reporting enrichments (both render-only — the data already lived on `GremlinResult`):

### #414 — Selected-test column in the HTML report
Adds a "Selected test" column to the HTML results table showing `killing_test` when present, else the joined `selected_tests`, else an em dash. Lets a reader see *which* test was run against each mutant directly in the report.

### #337 — Error diagnostics in the SonarQube export
`SonarQubeExporter` now exports errored gremlins (status `ERROR` with non-empty `error_output`) as `CODE_SMELL` issues (rule `mutant-error-{operator}`), message truncated to 500 chars — matching Stryker's existing truncation. Errored results with empty `error_output` are still excluded. Investigation confirmed the console, HTML, JSON, and Stryker formats already surfaced `error_output`; SonarQube was the only gap.

## Testing

- `DescribeHtmlReporterSelectedTests` (HTML column) and `DescribeSonarQubeExporterErrorOutput` (Sonar issues, empty-output exclusion, 500-char truncation) added.
- Adversarial QA gauntlet ran 9 probes across HTML escaping, empty/None handling, Sonar truncation boundaries (500/501, multi-byte), and Sonar schema conformance — **0 bugs found**; the new HTML cell uses the same `_escape_html` as sibling cells, truncation slices by code points, and the Sonar error issue is schema-identical to the survived-mutant issue.
- `uv run pytest tests -m small` → 1066 passed; mypy clean; ruff clean.

## Scope note

Issue #337's Gherkin references a `--gremlin-error-detail=N` CLI flag that does not exist in the codebase (the console reporter always shows errors when present). That flag is out of scope here and can be a follow-up if still desired.

Closes #414
Closes #337

### DIFF
--- a/docs/guide/reports.md
+++ b/docs/guide/reports.md
@@ -128,6 +128,7 @@ The HTML report includes:
 | Operator | Operator that created the gremlin |
 | Description | Human-readable mutation description |
 | Status | `zapped`, `survived`, `timeout`, or `error` |
+| Selected test | The test that killed the gremlin, or every test the coverage-based selector ran, or `—` if none |
 
 **Status Color Coding:**
 
@@ -180,6 +181,7 @@ The HTML report includes:
                     <th>Operator</th>
                     <th>Description</th>
                     <th>Status</th>
+                    <th>Selected test</th>
                 </tr>
             </thead>
             <tbody>
@@ -189,6 +191,7 @@ The HTML report includes:
                     <td>comparison</td>
                     <td>>= -> ></td>
                     <td class="status-survived">survived</td>
+                    <td>tests/test_auth.py::test_login_validates_age</td>
                 </tr>
                 <!-- More rows... -->
             </tbody>
--- a/docs/guide/reports.md
+++ b/docs/guide/reports.md
@@ -743,7 +743,7 @@ After uploading to Stryker Dashboard, add this badge to your README:
 
 ### SonarQube Export
 
-SonarQube can import surviving mutants as [external issues](https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/importing-external-issues/generic-issue-import-format/).
+SonarQube can import surviving mutants and errored gremlins as [external issues](https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/importing-external-issues/generic-issue-import-format/).
 
 #### Using SonarQubeExporter
 
@@ -777,15 +777,18 @@ sonar-scanner \
 
 #### What Gets Imported
 
-Only **survived** mutants are imported as issues:
+**Survived** mutants are always imported as issues. **Errored** gremlins are also
+imported, but only when they captured diagnostic output (`error_output`), so you
+can see why a gremlin failed to run rather than whether it survived:
 
-| Field | Value |
-|-------|-------|
-| Engine ID | `pytest-gremlins` |
-| Rule ID | `mutant-survived-{operator}` |
-| Severity | `MAJOR` (configurable) |
-| Type | `CODE_SMELL` |
-| Effort | `10 minutes` (configurable) |
+| Field | Survived | Errored |
+|-------|----------|---------|
+| Engine ID | `pytest-gremlins` | `pytest-gremlins` |
+| Rule ID | `mutant-survived-{operator}` | `mutant-error-{operator}` |
+| Severity | `MAJOR` (configurable) | `MAJOR` (configurable) |
+| Type | `CODE_SMELL` | `CODE_SMELL` |
+| Effort | `10 minutes` (configurable) | `10 minutes` (configurable) |
+| Message | `Mutant survived: {description}` | `Mutant errored: {error_output}` (truncated to 500 chars) |
 
 #### GitHub Actions for SonarQube
 
--- a/src/pytest_gremlins/reporting/html.py
+++ b/src/pytest_gremlins/reporting/html.py
@@ -890,6 +890,7 @@ class HtmlReporter:
                     <th scope="col">Operator</th>
                     <th scope="col">Description</th>
                     <th scope="col">Status</th>
+                    <th scope="col">Selected test</th>
                 </tr>
             </thead>
             <tbody>
@@ -903,6 +904,7 @@ class HtmlReporter:
         gremlin = result.gremlin
         status_class = f'status-{result.status.value}'
         diff_html = self._render_diff_panel(gremlin)
+        selected_test_html = self._render_selected_test(result)
         return f"""
                 <tr>
                     <td>
@@ -913,7 +915,25 @@ class HtmlReporter:
                     <td>{self._escape_html(gremlin.operator_name)}</td>
                     <td>{self._escape_html(gremlin.description)}</td>
                     <td class="{status_class}">{result.status.value}</td>
+                    <td>{selected_test_html}</td>
                 </tr>"""
+
+    def _render_selected_test(self, result: GremlinResult) -> str:
+        """Render the killing test, or the list of selected tests, for a result.
+
+        Args:
+            result: The GremlinResult to extract test information from.
+
+        Returns:
+            HTML-escaped killing test name when present; otherwise a
+            comma-separated, HTML-escaped list of selected test node IDs;
+            an em dash when neither is available.
+        """
+        if result.killing_test:
+            return self._escape_html(result.killing_test)
+        if result.selected_tests:
+            return self._escape_html(', '.join(result.selected_tests))
+        return '—'
 
     def _render_diff_panel(self, gremlin: Gremlin) -> str:
         """Render a collapsible side-by-side diff panel for a gremlin.
--- a/src/pytest_gremlins/reporting/sonarqube_export.py
+++ b/src/pytest_gremlins/reporting/sonarqube_export.py
@@ -1,7 +1,8 @@
 """SonarQube generic issue exporter for mutation testing results.
 
-Exports surviving mutants as external issues that can be imported into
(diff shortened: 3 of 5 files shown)


mikelane (commented):
## Review Summary — APPROVE (posted as COMMENT; solo-dev PR, GitHub blocks self-approval — see verdict artifact for the authoritative gate)

Reviewed at head SHA `95c442039d6943e71293879b25fe900008a94f90` (post update-branch; branch's merge-base with `origin/main` equals `origin/main` HEAD, so no regression risk from the merge — nothing on main to check against, confirming this is not a stale-branch situation).

### Scope
Exactly 5 files changed: `src/pytest_gremlins/reporting/html.py`, `src/pytest_gremlins/reporting/sonarqube_export.py`, `tests/reporting/test_html.py`, `tests/reporting/test_sonarqube_export.py`, `docs/guide/reports.md`. No scope creep from the update-branch merge commit (its diff is empty).

### #414 — HTML "Selected test" column
`_render_selected_test` (html.py) correctly prioritizes `killing_test` → joined `selected_tests` → em dash, and reuses the same `_escape_html` helper as every sibling cell in the row — no bespoke escaping path to audit separately. The a11y test (`scope_col_count == 6`) was correctly bumped alongside the new `<th>`.

### #337 — SonarQube error diagnostics
`_is_exportable` gates correctly: `SURVIVED` always exports; `ERROR` exports only when `error_output` is non-empty (preserves the existing "no noise from empty errors" behavior). `_build_error_issue`'s schema is field-for-field identical to `_build_survived_issue` (engineId/ruleId/severity/type/effortMinutes/primaryLocation), and the 500-char truncation slices the Python string directly (code-point safe, no encoding hazard).

### Tests
Both new `Describe` classes assert against public output (rendered HTML, parsed JSON) via the shared `make_result` factory — no reach into private state.

### Verified independently (not just trusting the PR description or gauntlet artifact)
- `uv run pytest tests/reporting/test_html.py tests/reporting/test_sonarqube_export.py -q` → 127 passed
- `ruff check` on the 4 changed Python files → clean
- `mypy` on the 2 changed src files → clean
- Confirmed via grep: no `--gremlin-error-detail` flag exists anywhere in `src/`, and `console.py` already calls `score.top_errors(limit=10)` unconditionally — the PR's scope note that this flag is a separable, non-existent follow-up (not a regression or omission) checks out.
- Confirmed `error_output` already flows through `results.py`, `console.py`, `json_reporter.py`, `stryker_export.py`, and the parallel pool/aggregator modules — SonarQube genuinely was the only format missing it.

### Non-blocking note (LOW)
`_build_survived_issue` and `_build_error_issue` build near-identical dict literals (differ only in `ruleId` prefix and `message` template). A shared private builder taking those two strings as params would remove the duplication, but at ~15 lines each this is a minor nit, not a SOLID violation worth blocking on.

**Decision: APPROVE**

PR #459, opened 2026-07-11
test+fix: coverage addopts follow-ups from #436 review (#444/#445/#446)

### LINKED ISSUE #444: test: cover --addopts final-token boundary in _addopts_without_cov (IndexError guard)
## Context

Follow-up from the #436 review (fix for #424). The new `_addopts_without_cov` parser in
`src/pytest_gremlins/plugin.py` has a `while index < count` guard that protects against a
pytest-cov option appearing as the **final token** with no following value (e.g. an `addopts`
ending in a bare `--cov`, or `--cov-report` with nothing after it).

The guard is correct, but no test exercises it. A mutant that flips `index < count` to
`index <= count` survives the entire suite, and the real failure mode is an `IndexError` raised
in production when a project's `addopts` ends in a value-taking cov option.

This was flagged HIGH by the test-coverage gauntlet specialist during the #436 review — it is the
single most dangerous untested path in that change.

## Acceptance

Add unit tests to `tests/plugin/test_plugin_addopts_preserve.py` covering:

- `_addopts_without_cov(['--cov'])` → `''` (bare cov as the only/last token, no IndexError)
- `_addopts_without_cov(['--strict-markers', '--cov-report'])` → `'--strict-markers'` (value-taking
  cov option as the final token)

Both must fail if the `index < count` boundary guard is removed/weakened (verify by mutating it
locally before committing).

## Notes

The implementation is correct today — this is a regression-locking test gap, not a live bug.

### LINKED ISSUE #445: test: end-to-end test that pytest honors re-injected -o addopts= (lock #424 behavior)
## Context

Follow-up from the #436 review (fix for #424). The fix re-injects the project's cov-stripped
`addopts` into gremlins' subprocess runs as `-o addopts=<...>`, so collection-affecting options
like `--import-mode=importlib` survive and coverage-guided selection stays engaged.

All tests in `tests/plugin/test_plugin_addopts_preserve.py` assert the **constructed command
string** (`addopts=--import-mode=importlib` is present in the argv). None runs a real pytest
subprocess to prove that pytest actually *honors* the re-injected `-o addopts=` value and that
`--import-mode=importlib` survives pytest's own `shlex.split` of the ini value.

The behavior is verified correct — during review, the bash-quoting gauntlet specialist round-tripped
the value through a live pytest subprocess and confirmed `shlex.quote`→`shlex.split` is lossless and
the option survives. But that proof lives in the review transcript, not in the committed suite. The
exact #424 regression (collection options silently dropped at the subprocess boundary) would NOT be
caught by any current test — they'd all stay green if the round-trip mangled the option.

## Acceptance

Add one integration/end-to-end test that:

- Builds a tiny throwaway project whose `addopts` includes `--import-mode=importlib`
- Runs a real pytest subprocess via the gremlins code path (or a faithful equivalent)
- Asserts the effect — coverage-guided selection re-engages (or, at minimum, pytest collects under
  importlib mode) — proving the re-injected `-o addopts=` is honored end-to-end

This regression-locks the behavior #424 depends on.

## Notes

Flagged HIGH (test-coverage purist) and MEDIUM (python purist, "wiring gap") in the #436 review.
Pairs with the value-consumption-heuristic robustness follow-up.

### LINKED ISSUE #446: robustness: harden _addopts_without_cov value-consumption heuristic against silent mis-strip
## Context

Follow-up from the #436 review (fix for #424). `_addopts_without_cov` in
`src/pytest_gremlins/plugin.py` decides whether a space-separated token following a pytest-cov
option is that option's *value* (drop it) or an unrelated token (keep it), using a heuristic:
"value-taking unless the option is in `_COV_FLAG_ONLY_OPTS`, unless the next token starts with `-`."

The heuristic matches pytest-cov's own greedy parse in the common case, but it can silently guess
wrong, and nothing detects when it does. Two cases the python-purist gauntlet specialist flagged:

1. **`--cov` has an *optional* value (`nargs='?'`)**: `--cov tests/` — pytest-cov may treat `tests/`
   as a positional, but the heuristic drops it as `--cov`'s value. Conversely a bare `--cov` meant
   to enable coverage-of-everything followed by a real positional could lose the positional.
2. **A cov value that legitimately starts with `-`** is NOT consumed (the `startswith('-')` guard),
   so it leaks into the subprocess as a standalone argv token — which can make the inner pytest
   error and the coverage scan silently return `{}`. That is the same silent-disengage failure
   class as the original #424 bug.

## Why it matters

A wrong guess degrades back into "coverage-guided selection silently disengages" — the exact
problem #424 set out to fix. Today it's correct for the common pytest-cov invocations, but it's an
unguarded mirror of an external tool's argument grammar.

## Possible directions (decide during implementation)

- Detect-and-warn when the heuristic is in an ambiguous branch (log that addopts may be
  mis-stripped, rather than silently guessing).
- Narrow the value-consumption to the known value-taking cov options only (explicit allowlist on
  both sides), treating unknowns conservatively.
- Add unit tests pinning the current behavior for `--cov <positional>`, `--cov-report -x`, and
  repeated value-taking cov options in sequence.

## Notes

Flagged MEDIUM (python purist) in the #436 review. The implementation is acceptable for current
pytest-cov; this hardens it against the silent-mis-strip failure mode.

mikelane:
## Summary

The three follow-ups from the #436 coverage-addopts review, closing the test gaps and hardening the value-consumption heuristic.

### #444 — final-token boundary tests
Cover the `index < count` guard in `_addopts_without_cov` (a value-taking cov option as the last token). The guard was correct but untested/mutant-survivable; now covered by `it_drops_a_bare_cov_that_is_the_final_token` and `it_drops_a_value_taking_cov_option_that_is_the_final_token`.

### #445 — end-to-end verification
A real subprocess test proving pytest genuinely honors the re-injected `-o addopts=<...>` at runtime — uses a duplicate-basename project that only collects under `--import-mode=importlib`, plus a negative control. Previous tests stubbed the subprocess, so this closes the gap where nothing proved pytest respected the re-injection.

### #446 — harden the value-consumption heuristic (production fix)
Fixes a **silent mis-strip**: dash-prefixed values for required-value cov options (`--cov-fail-under -1`, `--cov-report -x`) were leaking into the re-injected addopts string. Introduces a three-way classification — optional-value (`--cov`, `nargs='?'`), flag-only (`_COV_FLAG_ONLY_OPTS`), and required-value (everything else) — matching pytest-cov's actual option semantics.

## Testing

- 26 tests in `test_plugin_addopts_preserve.py` (incl. 3 new RED→GREEN for #446).
- Adversarial QA introspected the installed pytest-cov 7.1.0 option table and ran ~25 empirical probes across classification boundaries, dash/negative values, interleaving, prefix-over-strip, and shlex round-trip — **0 bugs**. Confirmed the over-strip guard is exact-match (`--covariance-threshold` survives) and classification matches pytest-cov exactly.
- `uv run pytest tests -m small` → 1067 passed; `-m "small or medium"` → 1507 passed; mypy + ruff clean.

Closes #444
Closes #445
Closes #446

### DIFF
--- a/src/pytest_gremlins/plugin.py
+++ b/src/pytest_gremlins/plugin.py
@@ -1787,6 +1787,16 @@ _COV_FLAG_ONLY_OPTS = frozenset(
     }
 )
 
+# ``--cov`` alone takes an *optional* value (pytest-cov defines it with
+# ``nargs='?'``): a bare ``--cov`` immediately followed by another option has no
+# value to consume, so the dash-prefix heuristic below only applies to this one
+# option. Every other value-taking cov option (``--cov-report``,
+# ``--cov-fail-under``, ``--cov-precision``, etc.) *requires* its value and must
+# consume the next token unconditionally -- even one that happens to start with
+# ``-`` (e.g. ``--cov-fail-under -1``) -- or it silently leaks into the
+# re-injected addopts string as a stray, unrelated-looking token (issue #446).
+_COV_OPTIONAL_VALUE_OPTS = frozenset({'--cov'})
+
 
 def _is_cov_addopt(token: str) -> bool:
     """Return True if ``token`` is a pytest-cov option.
@@ -1812,8 +1822,11 @@ def _addopts_without_cov(raw_addopts: list[str]) -> str:
 
     Only the ``--cov*`` / ``--no-cov*`` option family is stripped. A cov option written
     with a space-separated value (``--cov-report term``, ``--cov-precision 2``) has its
-    value token dropped as well; the known value-less switches in
-    ``_COV_FLAG_ONLY_OPTS`` keep their following token, and inline forms
+    value token dropped as well: required-value options (everything except ``--cov``
+    itself and the known value-less switches in ``_COV_FLAG_ONLY_OPTS``) consume their
+    following token unconditionally, even one that starts with ``-``
+    (``--cov-fail-under -1``); ``--cov``'s optional value is only dropped when the
+    following token doesn't itself look like an option. Inline forms
     (``--cov-report=term``) are self-contained. The result is a single string suitable
     for ``-o addopts=<...>``.
 
@@ -1829,16 +1842,13 @@ def _addopts_without_cov(raw_addopts: list[str]) -> str:
         token = raw_addopts[index]
         index += 1
         if _is_cov_addopt(token):
-            # A value-taking cov option written without an inline ``=`` consumes the
-            # next token as its value; drop that too — unless this is a known value-less
-            # switch or the next token is itself an option.
-            if (
-                '=' not in token
-                and token not in _COV_FLAG_ONLY_OPTS
-                and index < count
-                and not raw_addopts[index].startswith('-')
-            ):
-                index += 1
+            has_no_inline_value = '=' not in token
+            is_value_taking = token not in _COV_FLAG_ONLY_OPTS
+            has_next_token = index < count
+            if has_no_inline_value and is_value_taking and has_next_token:
+                next_looks_like_option = raw_addopts[index].startswith('-')
+                if token not in _COV_OPTIONAL_VALUE_OPTS or not next_looks_like_option:
+                    index += 1
             continue
         kept.append(token)
     return ' '.join(shlex.quote(token) for token in kept)
--- a/tests/plugin/test_plugin_addopts_preserve.py
+++ b/tests/plugin/test_plugin_addopts_preserve.py
@@ -13,6 +13,7 @@ See: https://github.com/mikelane/pytest-gremlins/issues/113
 from __future__ import annotations
 
 from pathlib import Path
+import subprocess
 from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


mikelane (commented):
## PR Review — APPROVE

**Scope**: `src/pytest_gremlins/plugin.py` (34 lines) + `tests/plugin/test_plugin_addopts_preserve.py` (+80 lines). Confirmed no unrelated changes were dragged in by the update-branch merge.

### Regression check (squash-merge hazard, branch age)
`git merge-base HEAD origin/main` shows **zero commits on `main` touched `plugin.py`** since divergence. Main's only changes since divergence are in `reporting/html.py`, `reporting/sonarqube_export.py`, and docs — no overlap with this PR's files. Deletions (20) vs additions (106) confirm this is net-additive hardening, not a wipeout. No stale-branch regression risk.

### #446 — three-way classification
`_COV_OPTIONAL_VALUE_OPTS = {'--cov'}` (nargs='?'), `_COV_FLAG_ONLY_OPTS` (store_true/const), and required-value fall-through are correct and readable. The branch logic uses named booleans (`has_no_inline_value`, `is_value_taking`, `has_next_token`, `next_looks_like_option`) instead of a dense nested conditional — self-documenting, no WHAT-comments needed. `_is_cov_addopt` matching is exact-name + hyphenated-family, not prefix, so `--covariance-threshold` is correctly preserved (verified by reading the function and by the gauntlet's empirical probes against pytest-cov 7.1.0's real option table).

### Tests
26 tests, all passing locally (`uv run pytest tests/plugin/test_plugin_addopts_preserve.py -v`). Declarative `it_...` names, one behavior per test, no private-method access. The #445 e2e test (`DescribeReinjectedAddoptsHonoredByRealPytest`) spawns a real pytest subprocess against a throwaway project with duplicate test basenames (a case only `--import-mode=importlib` can collect), with both a positive assertion and a negative control — genuine proof the re-injected addopt takes effect, not a stubbed assertion.

### Quality gates
`ruff check` clean, `ruff format --check` clean, `mypy` (strict) clean. CI: Lint/Type Check/Docs/Security Audit green; Python test matrix was still in progress at review time (not failing) — merge-guardian should confirm completion before merge.

**Findings**: 0 CRITICAL, 0 HIGH, 0 MEDIUM, 0 LOW. No blockers.

PR #465, opened 2026-07-15
fix(cache): hash file contents as bytes

### LINKED ISSUE #455: [Bug]: ContentHasher.hash_file crashes on PEP 263 non-utf-8 source (cache/hasher.py:53)

Sanjays2402:
## Summary

Hash incremental cache keys from raw bytes so valid non-UTF-8 Python source no longer raises `UnicodeDecodeError`. Fixes #455.

## Changes

- Hash file bytes directly with SHA-256.
- Cover PEP 263 Latin-1 source and arbitrary binary content.

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## Related Issues

Fixes #455

## Testing

- [ ] All existing tests pass
- [x] Added new tests for this change
- [x] Manually tested

The regression fails with `UnicodeDecodeError` before the fix and passes after it. All 99 cache tests, Ruff, formatting, and mypy pass. The full suite reaches 1,491 passes; its 32 process-pool failures are identical on the baseline under the network/process-denied sandbox.

## Checklist

- [x] Tests written FIRST (TDD)

### DIFF
--- a/src/pytest_gremlins/cache/hasher.py
+++ b/src/pytest_gremlins/cache/hasher.py
@@ -56,8 +56,7 @@ class ContentHasher:
         Raises:
             FileNotFoundError: If the file does not exist.
         """
-        content = path.read_text(encoding='utf-8')
-        return self.hash_string(content)
+        return hashlib.sha256(path.read_bytes()).hexdigest()
 
     def hash_files(self, paths: list[Path]) -> dict[str, str]:
         """Hash multiple files and return a mapping of path to hash.
--- a/tests/cache/test_bug_hunting.py
+++ b/tests/cache/test_bug_hunting.py
@@ -4,6 +4,8 @@ These tests expose potential bugs found during code review.
 Following TDD: write failing test first, then fix the bug.
 """
 
+import hashlib
+
 import pytest
 
 from pytest_gremlins.cache.hasher import ContentHasher
@@ -35,19 +37,14 @@ class DescribeHasherBugs:
         # Currently they're the same, which is a bug
         assert combined_empty == string_empty  # This passes, showing the bug exists
 
-    def it_raises_clear_error_for_binary_file_content(self, tmp_path):
-        """hash_file raises UnicodeDecodeError for binary files.
-
-        BUG: hash_file uses read_text() which fails on binary files.
-        The error message is not helpful for users.
-        """
+    def it_hashes_binary_file_content(self, tmp_path):
+        """hash_file hashes binary files without decoding them."""
         hasher = ContentHasher()
         binary_file = tmp_path / 'test.pyc'
-        binary_file.write_bytes(b'\x00\x01\x02\x03\xff\xfe')
+        content = b'\x00\x01\x02\x03\xff\xfe'
+        binary_file.write_bytes(content)
 
-        # Currently raises UnicodeDecodeError - not very helpful
-        with pytest.raises(UnicodeDecodeError):
-            hasher.hash_file(binary_file)
+        assert hasher.hash_file(binary_file) == hashlib.sha256(content).hexdigest()
 
 
 @pytest.mark.medium
--- a/tests/cache/test_hasher.py
+++ b/tests/cache/test_hasher.py
@@ -144,6 +144,15 @@ class DescribeContentHasherFileIO:
 
         assert hasher.hash_file(lf_path) != hasher.hash_file(crlf_path)
 
+    def it_hashes_non_utf8_source_bytes(self, tmp_path):
+        """hash_file hashes valid PEP 263 source without decoding it."""
+        hasher = ContentHasher()
+        content = b"# -*- coding: latin-1 -*-\nname = 'caf\xe9'\n"
+        file_path = tmp_path / 'latin1.py'
+        file_path.write_bytes(content)
+
+        assert hasher.hash_file(file_path) == hashlib.sha256(content).hexdigest()
+
     def it_raises_for_missing_file(self, tmp_path):
         """hash_file raises FileNotFoundError for missing files."""
         hasher = ContentHasher()


mikelane:
Thanks for this, @Sanjays2402 — the fix is right and the tests are well done. Hashing raw bytes is the correct approach for a cache key, and I particularly liked that you *inverted* the old `it_raises_clear_error_for_binary_file_content` test rather than deleting it. That test was asserting the bug; turning it into an assertion of the fixed behavior is exactly right.

Unfortunately I need to ask you for a rebase, and the conflict is my fault rather than yours.

## What happened

Your change surfaced a couple of adjacent problems, and I fixed them in #473, which merged a few minutes ago. One of those fixes touched `tests/cache/test_hasher.py` — the same file you edited — so GitHub can no longer auto-merge your branch. Had I merged yours first, this wouldn't have come up.

For context, what #473 changed:

- **`it_matches_hash_file_output_to_hash_string` was renamed** to `it_matches_hash_file_output_to_hash_string_for_lf_source` and now writes explicit LF bytes. It was using `write_text()`, which emits the platform's default newline — so on Windows it wrote CRLF, and once your change lands it would have compared a CRLF-byte digest against an LF-string digest and failed. Your PR was correct; that test was quietly platform-dependent and nobody had noticed.
- **Two `xfail` tests were added** that pin the behavior your PR introduces. They fail today on purpose and will pass once yours merges.
- Unrelated fixes to Windows CI and a `pygments` advisory rode along in the same PR.

## What I need

Just the conflict resolution — nothing else.

```bash
git fetch upstream main          # or whatever you've named the mikelane remote
git rebase upstream/main
# resolve tests/cache/test_hasher.py
git push --force-with-lease
```

If you'd rather avoid a force-push, merging works just as well and I'll squash it on the way in:

```bash
git fetch upstream main
git merge upstream/main
# resolve tests/cache/test_hasher.py
git push
```

**The resolution is "keep both sides."** The conflict is purely additive — your `it_hashes_non_utf8_source_bytes` and the new `xfail` tests were both appended to `DescribeContentHasherFileIO` at the same spot. Nothing is actually in tension; git just can't tell how to interleave them.

## What you don't need to do

Leave the `xfail` markers alone. Your change is what makes those tests pass, so they'll flip to `xpass` when this merges — removing them is follow-up work I'll handle on my side. Please don't feel obliged to take that on.

If the rebase turns into a hassle or you'd rather not deal with it, say so and I'll sort it out another way — you've already done the hard part here, and I don't want a merge-order mistake of mine to turn into a chore for you. 🙏

Sanjays2402:
Rebased onto main — the conflict's resolved and it's mergeable again. No worries about the ordering, these things happen.

The conflict was just the two adjacent test blocks in test_hasher.py, so I kept both sides in full: your LF rename (it_matches_hash_file_output_to_hash_string_for_lf_source with explicit LF bytes) and your two xfail tests from #473, alongside my byte-hashing assertion and the inverted binary-content test. Since my fix lands on this branch, those two xfails now xpass — and because they're strict=False (and there's no xfail_strict in pyproject) that's a clean pass, not a failure, so I left the markers as-is rather than stripping them.

Full suite is green: 1522 passed, 2 xpassed, 0 failed. Ready when you are.

mikelane:
Perfect, thank you — the rebase is clean and your read on the `xfail` markers is exactly right. Since there's no `xfail_strict` in `pyproject.toml`, those two xpass rather than fail, so leaving them was the correct call on your end. They're our tests (added in #473), so removing them now that your fix makes them redundant is our follow-up to handle, not yours.

Running it through the review gate now. Appreciate you sticking with this through the rebase — the fix is a good one, and thanks for the patience with a conflict our own merge order created. 🙏

mikelane (approved):
## Review — PR #465 (rebased, SHA 2791885)

Verified independently (not rubber-stamping the prior adversarial-qa pass):

**Diff scope** — confirmed exactly three files changed, matching the description: `src/pytest_gremlins/cache/hasher.py` (+1/-2), `tests/cache/test_bug_hunting.py` (+7/-10), `tests/cache/test_hasher.py` (+9/-0). No rebase-introduced extras.

**Correctness of the fix** — `hash_file` now does `hashlib.sha256(path.read_bytes()).hexdigest()` instead of decoding through `hash_string(path.read_text(encoding='utf-8'))`. This is the right fix for a content-addressable cache key: hashing raw bytes eliminates the `UnicodeDecodeError` on non-UTF-8 source (the PEP 263 latin-1 case from #455) and on binary files, without any regression risk on the happy path — the docstring already scopes hash_file/hash_string equivalence to LF-encoded UTF-8 text only (pre-existing, from #469), so this change doesn't break a promise that was never made. Checked every caller (`plugin.py:928`, two perf test files) — all treat the return value as an opaque cache key, none cross-compare against `hash_string` output or depend on the old `UnicodeDecodeError` behavior. No false cache HIT is introduced: a cross-scheme collision between old-text-hash and new-byte-hash would require a SHA-256 break; same-scheme keys only collide on byte-identical content.

**Test quality** — `it_hashes_binary_file_content` (renamed from the old bug-reproduction test) independently recomputes `hashlib.sha256(content).hexdigest()` rather than mirroring the production call, so it's a real behavioral assertion, not a tautology — it would catch a revert to `read_text`. Same for the new `it_hashes_non_utf8_source_bytes` latin-1 case.

**xfail → xpass** — the two `#473`-tagged xfail markers in `test_hasher.py` (raw-bytes hashing, CRLF/LF distinction) now xpass. Ran the suite locally: `99 passed, 0 failed, 2 xpassed` — confirms `strict=False` is honored (no project-wide `xfail_strict` override exists). This is a clean pass, not a silent failure. LOW finding: an always-xpassing xfail gives zero regression protection going forward — flagging as a non-blocking follow-up to remove the markers now that #465 has landed (the in-code comments already say as much but don't reference a tracking issue).

**Regression check (squash-merge hazard)** — `git merge-base HEAD origin/main` equals `origin/main`'s current head (`b5b126a`) exactly. The branch isn't just "not stale" — it's rebased directly onto current main, so there is zero risk of this squash-merge reverting any sibling-PR work. Diff shape (17 additions / 12 deletions) is unremarkable for a feature/bugfix PR.

**Tooling** — ruff check: clean. ruff format --check: clean. mypy (strict): no issues.

**Verdict: APPROVE.** 0 CRITICAL, 0 HIGH, 0 MEDIUM, 1 LOW (non-blocking). Nice first contribution — the fix is minimal, correct, and the test rewrite properly asserts the new behavior instead of just deleting the old assertion.

Note: `mergeStateStatus: BLOCKED` on this PR reflects the missing approving review this comment provides, not a CI or merge-conflict issue — merge-guardian's separate CI/merge-state gate is unaffected by this review.

mikelane:
Merged — thank you, @Sanjays2402! 🎉

This is a real fix for a real crash (`UnicodeDecodeError` on valid non-UTF-8 source), and hashing raw bytes is exactly the right approach for a content cache key. Nicely done for a first contribution — clean TDD, you inverted the old bug-asserting test rather than deleting it, and you handled the rebase gracefully when our own merge order forced one.

Closes #455. The two `xfail` markers we added in #473 now pass because of your change; we'll clear those on our side. Thanks again for the patience and the quality work.

PR #472, opened 2026-07-20
chore(deps): bump pillow and click to clear pip-audit CVEs (closes #470)

mikelane:
## Summary

`Security Audit` is a **required** status check on `main` and was failing on every open PR, blocking all merges. `pip-audit` reported 9 unignored advisories:

| package | from | to | advisories |
| --- | --- | --- | --- |
| `click` | 8.3.1 | 8.4.2 | PYSEC-2026-2132 |
| `pillow` | 12.2.0 | 12.3.0 | PYSEC-2026-2253/2254/2255/2256/2257/3451/3452/3453 |

No code change caused this — the live advisory database retroactively failed a previously-green pipeline.

## Scope

`uv.lock` only. Exactly two packages changed, zero added, zero removed (verified by parsing both lockfile revisions with `tomllib`; `uv lock --check` exits 0).

## Runtime impact: none

Verified, not assumed. `pyproject.toml` runtime `dependencies` are exactly `coverage` and `pytest`. `click` is reachable only via `mkdocs`; `pillow` only via `mkdocs-material[imaging]` and `cairosvg` — both gated behind the `docs` extra. **The published wheel is unaffected.** A user installing `pytest-gremlins[docs]` receives both, and is helped by the bump rather than harmed.

## Verification

- `uv run pip-audit --ignore-vuln CVE-2026-4539` → `No known vulnerabilities found, 1 ignored`
- `uv run pytest tests -m "small or medium and not wip"` → **1522 passed, 0 failed, 2 skipped**
- `uv run mypy src/pytest_gremlins` → no issues
- `uv run ruff check src tests` → clean
- `uv run mkdocs build --strict` → passes, no escalated warnings

`click` resolved to 8.4.2, a minor bump beyond the 8.3.3 the advisory required, so behavior risk was checked explicitly: there is zero direct `click` or `PIL` usage anywhere in the repo, and the strict docs build passes.

## Adjacent finding, deliberately not fixed here

The adversarial QA gate found that `pygments` PYSEC-2026-2987 is suppressed at `ci.yml:180` by a stale `# pygments: no fix available yet` comment — 2.20.0 now exists. Unlike click/pillow, `pygments` is a runtime dependency of `pytest` and **does** ship to users. Filed as #471 and scoped to the follow-up PR rather than folded in here, so this branch stays atomic and unblocks the queue immediately.

## Also noted (INFO, non-blocking)

[...]

PR #473, opened 2026-07-20
fix: clear the three release blockers — hasher test, Windows CI integrity, pygments CVE

### LINKED ISSUE #468: ci: windows-latest jobs cannot fail — pwsh discards pytest exit code
## Summary

**The entire four-job `windows-latest` test matrix cannot fail.** Test failures on Windows are silently discarded and the job reports `success`.

This is pre-existing (not introduced by any current PR) and `release.yml` has the identical pattern, so **release gating is affected too**.

## Evidence

Run [29712569172](https://github.com/mikelane/pytest-gremlins/actions/runs/29712569172), job `88258923762` — `Test (Python 3.12, windows-latest)`:

- GitHub reports the job conclusion as **`success`**, and the PR check rollup renders it as **pass**
- Its own log contains:

```
FAILED tests/cache/test_hasher.py::DescribeContentHasherFileIO::it_matches_hash_file_output_to_hash_string [MEDIUM]
    - AssertionError: assert 'eb35a385f168...' == '961edaeae954...'
===== 1 failed, 1522 passed, 2 skipped, 1 deselected in 196.67s =====
```

All four windows jobs (3.11, 3.12, 3.13, 3.14) show the same pattern.

## Root cause

`.github/workflows/ci.yml:114-122`:

```yaml
      - name: Run tests with coverage
        env:
          COVERAGE_PROCESS_START: pyproject.toml
          PYTHONPATH: .
        run: |
          uv run coverage run -m pytest tests -m "small or medium and not wip" -v
          uv run coverage combine --quiet || echo "::warning::No coverage data to combine (expected in some environments)"
          uv run coverage report --show-missing
          uv run coverage xml
```

A multi-line `run:` block with **no `shell:` override**. On `windows-latest`, GitHub Actions defaults to PowerShell — confirmed in the job log:

```
shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
```

PowerShell has no `set -e` equivalent for multi-statement scripts. **The step's exit code is that of the last command** — `uv run coverage xml`. The non-zero exit from the `pytest` line is discarded.

(The `|| echo` on the `coverage combine` line is bash idiom, which suggests the whole block was written assuming bash.)

## Impact

- Every Windows CI result on every PR since this workflow was written has been non-load-bearing.
- Windows-only regressions can reach `main` with a green check.
- `release.yml:41` shares the pattern, so a release can be cut on unverified Windows results.

This is the `|| true` failure mode from `~/.claude/rules/release-hygiene.md` — a masked failure — arriving via shell selection rather than an explicit suppression.

## Fix

Add an explicit shell to the step:

```yaml
      - name: Run tests with coverage
        shell: bash
        env:
          ...
```

Or set it workflow-wide:

```yaml
defaults:
  run:
    shell: bash
```

Then apply the same fix to `release.yml:41`.

## Acceptance criteria

- [ ] `ci.yml` test step runs under an explicit shell with fail-fast semantics
- [ ] `release.yml:41` likewise
- [ ] Audit every other multi-line `run:` block in `.github/workflows/` for the same defect
- [ ] Verified: a deliberately failing test on Windows turns the job red
- [ ] Blocked-on: #469 (the failing test must be fixed first, or main goes red the moment this lands)

## Found by

Adversarial QA gate on PR #465. Worth noting it was found *outside* the PR's diff — the change under review was 3 files, +19/-12.

### LINKED ISSUE #469: test: re-scope it_matches_hash_file_output_to_hash_string for byte-exact hashing
## Summary

`tests/cache/test_hasher.py::DescribeContentHasherFileIO::it_matches_hash_file_output_to_hash_string` fails on Windows once PR #465 lands, because it asserts an invariant that has quietly become newline-conditional.

**The fix belongs in the test, not in production code.**

## Background

PR #465 (thanks @Sanjays2402) fixed #455 by changing `ContentHasher.hash_file`:

```python
# before
content = path.read_text(encoding='utf-8')
return self.hash_string(content)

# after
return hashlib.sha256(path.read_bytes()).hexdigest()
```

That change is correct — hashing raw bytes is the right way to derive a content hash, and it's what fixes the `UnicodeDecodeError` on PEP 263 non-UTF-8 source.

But `read_text()` applies **universal-newline translation** and `read_bytes()` does not. Verified empirically:

| | CRLF vs LF file, same logical content |
| --- | --- |
| old `hash_file` | hash **identically** |
| new `hash_file` | hash **differently** |

## The failing test

```python
def it_matches_hash_file_output_to_hash_string(self, tmp_path):
    ...
    assert file_hash == string_hash
```

It writes its fixture with `Path.write_text(content)`, which uses the **platform default newline**. On Linux/macOS that's LF and the assertion holds. On Windows it's CRLF, so `hash_file` sees CRLF bytes while `hash_string` sees LF — and the digests differ:

```
AssertionError: assert 'eb35a385f168...' == '961edaeae954...'
```

Confirmed failing on all four `windows-latest` jobs of run 29712569172.

It has been invisible because of the CI shell defect in #468 — Windows jobs report `success` regardless.

## Why this is our debt, not the contributor's

The `hash_file == hash_string` equivalence was **always** newline-dependent. It simply never mattered while both sides went through `read_text()`. PR #465 didn't introduce the fragility; it exposed it. The test was not re-scoped because nothing signalled that it needed to be.

Note the test still passes on Linux/macOS — so it now **passes for the wrong reason**, which reads as coverage it no longer provides. That's the part worth fixing carefully.

## Fix

1. Re-scope the existing test to write **explicit LF bytes** and rename it to reflect its actual (LF-only) scope — e.g. `it_matches_hash_file_output_to_hash_string_for_lf_source`.
2. Add a test that **pins the new behavior explicitly**: a CRLF file and an LF file with the same logical content hash *differently*, and `hash_file` equals `sha256` of the raw bytes.
3. Update `hash_file`'s docstring to state it hashes raw bytes, so equivalence with `hash_string` holds only for LF-encoded UTF-8.

**Do not** "fix" this by restoring newline normalization in `hash_file`. Byte-exact hashing is correct for a cache key: byte-differing files *should* produce different keys. Adversarial QA confirmed the new behavior is strictly less collision-prone, and that a CRLF/platform divergence produces a cache **miss** (safe recompute), never a false **hit**.

## Acceptance criteria

- [ ] Existing test writes explicit LF bytes and is named for its real scope
- [ ] A test pins CRLF-vs-LF divergence as intended behavior
- [ ] `hash_file` docstring documents raw-byte hashing and the LF-only equivalence
- [ ] `uv run pytest tests -m "small or medium"` green on Linux, macOS **and** Windows
- [ ] Production `hasher.py` behavior unchanged from #465

## Found by

Adversarial QA gate on PR #465.

### LINKED ISSUE #471: security: pygments CVE suppression is stale — fix exists and it ships to users
## Summary

`pygments` PYSEC-2026-2987 is suppressed in CI by a **stale** justification. A fix now exists, and unlike the other advisories we've been clearing, this one **ships to every user**.

## The suppression

`.github/workflows/ci.yml:180`:

```yaml
# pygments: no fix available yet
uv run pip-audit --ignore-vuln CVE-2026-4539
```

That comment was true when written. It isn't now — `pygments` 2.20.0 is released and fixes the advisory:

```
$ uv run pip-audit          # without the ignore flag
pygments 2.19.2  PYSEC-2026-2987  Fix Versions: 2.20.0
```

## Why this one is different

The pillow and click advisories cleared in #470 were **docs tooling** — reachable only via the `docs` extra, never in the published wheel's dependency closure.

`pygments` is not. It is a runtime dependency of `pytest`, and `pytest` is a base dependency of `pytest-gremlins`:

```toml
dependencies = [
    "coverage>=7.12.0",
    "pytest>=7.0.0",
]
```

So every user who installs `pytest-gremlins` receives `pygments`. This is the only advisory in the current set that reaches shipped code.

## Fix

```bash
uv lock --upgrade-package pygments
```

Verified during the #470 adversarial QA gate to resolve cleanly to 2.20.0 with **zero collateral lockfile changes** (then reverted, as it was out of scope for that branch).

Then remove the now-unnecessary suppression from `ci.yml:180` — both the `--ignore-vuln CVE-2026-4539` flag and the stale comment — so `pip-audit` runs unsuppressed and CI tells the truth again.

## Acceptance criteria

- [ ] `pygments` bumped to >= 2.20.0 in `uv.lock`
- [ ] `--ignore-vuln CVE-2026-4539` and its comment removed from `ci.yml`
- [ ] `uv run pip-audit` (no ignore flags) reports no vulnerabilities
- [ ] Test suite green

## Release gate

**This blocks the next release.** It's the only known advisory affecting the published artifact.

## Notes

Worth a broader lesson: a `--ignore-vuln` flag justified by "no fix available yet" is a claim with an expiry date, and nothing re-checks it. Consider either annotating suppressions with a review date, or dropping the flag entirely and letting the job fail loudly when an advisory lands — a failing Security Audit is more useful than a silently stale exception.

## Found by

Adversarial QA gate on the #470 lockfile bump — flagged as an adjacent finding while probing the dependency closure.

mikelane:
Clears all three known release blockers. Ordered deliberately, and the order matters.

## 1. `395dfe8` — re-scope the hasher test (closes #469)

`it_matches_hash_file_output_to_hash_string` wrote its fixture with `write_text()`, i.e. the platform-default newline. Under today's `read_text()`-based `hash_file` that passes everywhere, because universal-newline translation cancels it out. Once #465 lands and `hash_file` hashes raw bytes, the Windows CRLF fixture is compared against an LF string and the assertion fails.

Re-scoped to write explicit LF bytes and renamed to `..._for_lf_source` to state its real scope. Verified to pass **both** against current `read_text()` hashing and against #465's `read_bytes()` version — which is what makes landing this first safe.

Added two `xfail(strict=False)` tests pinning the post-#465 contract, plus a docstring-only note on `hash_file` that `hash_string` equivalence holds only for LF-encoded UTF-8. **`hasher.py` logic is unchanged** — the production fix is #465's to make.

## 2. `003fff1` — make windows-latest able to fail (closes #468)

The "Run tests with coverage" step is a multi-line `run:` with no `shell:` override. On Windows that runs under `pwsh -command`, which does not halt on a native command's non-zero exit — so a pytest failure was overwritten by the exit code of the final `coverage` command. **All four Windows jobs have reported `success` regardless of test outcome for this workflow's entire history**, and `release.yml` shared the defect, so releases were gated on the same fiction.

`shell: bash` maps to `bash --noprofile --norc -eo pipefail` on Windows, so `-e` aborts at the first failure. Applied to `ci.yml` and `release.yml`. Every other multi-line `run:` block across all six workflow files was audited and is pinned to `ubuntu-latest`, where bash is already the default — no other instance exists.

> **Correction:** commit `003fff1`'s message states this must land *before* #469. That is backwards — #469 must land first, or Windows becomes load-bearing before the fixed test is in place. The commit order in this branch is correct; only that message's rationale is inverted.

## 3. `9f9b743` — pygments CVE (closes #471)

`pygments` PYSEC-2026-2987 was suppressed by a `# pygments: no fix available yet` comment that expired when 2.20.0 shipped. Unlike the pillow/click bumps in #472 (docs tooling), **pygments is a runtime dependency of pytest and ships in every published wheel**. Bumped to 2.20.0 with zero collateral lockfile changes; the `--ignore-vuln` flag and its stale comment are removed so `pip-audit` runs unsuppressed.

## ⚠️ Expect the Windows matrix to be informative

This PR produces the **first trustworthy Windows CI result in this repository's history**. Adversarial QA flagged two surfaces that have never been validated by a job capable of failing:

- `fail_under = 97` now genuinely gates the Windows legs, and coverage can differ by platform
- Windows-specific timing thresholds in `test_pool_optimization_benefits.py:137-138` and `test_optimized_pool.py:104` were written defensively and never actually exercised

**If Windows goes red here, that is a pre-existing failure this PR reveals, not one it introduces.** It would still need fixing before release.

## Verification

- `pytest -m "small or medium and not wip"` → 1522 passed, 2 skipped, 2 xfailed
- doctests → 38 passed, 2 skipped
- `mypy` clean · `ruff check` clean · `ruff format --check` 189 files formatted
- `pip-audit` (no ignore flags) → no known vulnerabilities
- `uv lock --check` → clean, no drift

## Unblocks

PR #465 can then merge with its gauntlet passing honestly — and validated by a Windows matrix that can actually fail.

Closes #469
Closes #468
Closes #471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,7 @@ jobs:
         run: uv sync
 
       - name: Run tests with coverage
+        shell: bash
         env:
           COVERAGE_PROCESS_START: pyproject.toml
           PYTHONPATH: .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,7 +178,7 @@ jobs:
         run: uv sync
 
       - name: Run pip-audit
-        run: uv run pip-audit --ignore-vuln CVE-2026-4539  # pygments: no fix available yet
+        run: uv run pip-audit
 
   attrs-smoke:
     name: Attrs Smoke Test
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,6 +34,7 @@ jobs:
         run: uv sync
 
       - name: Run tests with coverage (small + medium, matching PR CI)
+        shell: bash
         env:
           COVERAGE_PROCESS_START: pyproject.toml
           PYTHONPATH: .
--- a/src/pytest_gremlins/cache/hasher.py
+++ b/src/pytest_gremlins/cache/hasher.py
@@ -41,6 +41,12 @@ class ContentHasher:
     def hash_file(self, path: Path) -> str:
         """Hash a file's content and return its hex digest.
 
+        Equivalence with :meth:`hash_string` for the same logical content
+        is guaranteed only when the file is LF-encoded UTF-8 text. Files
+        with CRLF line endings or non-UTF-8 encodings are not guaranteed
+        to produce the same hash as ``hash_string`` on the equivalent
+        string.
+
         Args:
             path: Path to the file to hash.
 
--- a/tests/cache/test_hasher.py
+++ b/tests/cache/test_hasher.py
@@ -4,6 +4,8 @@ Content hashing is the foundation of incremental analysis. Files with
 identical content produce identical hashes, enabling cache lookups.
 """
 
+import hashlib
+
 import pytest
 
 from pytest_gremlins.cache.hasher import ContentHasher
@@ -87,17 +89,60 @@ class DescribeContentHasherFileIO:
         assert isinstance(result, str)
         assert len(result) == 64  # SHA-256 produces 64 hex characters
 
-    def it_matches_hash_file_output_to_hash_string(self, tmp_path):
-        """hash_file produces same result as hash_string for same content."""
+    def it_matches_hash_file_output_to_hash_string_for_lf_source(self, tmp_path):
+        """hash_file matches hash_string for LF-encoded content.
+
+        This equivalence is scoped to LF-encoded UTF-8 source: it does not
+        claim anything about CRLF or non-UTF-8 files. See #469.
+        """
         hasher = ContentHasher()
         content = 'class MyClass:\n    pass\n'
         file_path = tmp_path / 'test.py'
-        file_path.write_text(content)
+        file_path.write_bytes(content.encode('utf-8'))
 
         file_hash = hasher.hash_file(file_path)
         string_hash = hasher.hash_string(content)
 
         assert file_hash == string_hash
+
+    @pytest.mark.xfail(
+        reason='passes once #465 lands: hash_file will hash raw bytes instead of read_text() output',
+        strict=False,
+    )
+    def it_hashes_file_content_as_raw_bytes(self, tmp_path):
+        """hash_file equals the SHA-256 digest of the file's raw bytes.
+
+        Fails on current main because hash_file still hashes read_text()
+        output, which applies universal-newline translation. Remove the
+        xfail marker once #465 merges.
+        """
+        hasher = ContentHasher()
+        file_path = tmp_path / 'test.py'
+        file_path.write_bytes(b'class MyClass:\r\n    pass\r\n')
+
+        result = hasher.hash_file(file_path)
+
+        assert result == hashlib.sha256(file_path.read_bytes()).hexdigest()
+
+    @pytest.mark.xfail(
+        reason='passes once #465 lands: hash_file will hash raw bytes instead of read_text() output',
+        strict=False,
+    )
+    def it_hashes_crlf_and_lf_files_differently_for_same_logical_content(self, tmp_path):
+        """A CRLF file and an LF file with the same logical content hash differently.
+
+        Fails on current main because read_text()'s universal-newline
+        translation makes CRLF and LF files hash identically. Remove the
+        xfail marker once #465 merges.
+        """
+        hasher = ContentHasher()
+        logical_content = 'class MyClass:\n    pass\n'
+        lf_path = tmp_path / 'lf.py'
+        crlf_path = tmp_path / 'crlf.py'
+        lf_path.write_bytes(logical_content.encode('utf-8'))
+        crlf_path.write_bytes(logical_content.replace('\n', '\r\n').encode('utf-8'))
+
+        assert hasher.hash_file(lf_path) != hasher.hash_file(crlf_path)
 
     def it_raises_for_missing_file(self, tmp_path):
         """hash_file raises FileNotFoundError for missing files."""
--- a/uv.lock
+++ b/uv.lock
@@ -1567,11 +1567,11 @@ wheels = [
 
 [[package]]
 name = "pygments"
-version = "2.19.2"
+version = "2.20.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/b0/77/a5b8c569bf593b0140bde72ea885a803b82086995367bf2037de0159d924/pygments-2.19.2.tar.gz", hash = "sha256:636cb2477cec7f8952536970bc533bc43743542f70392ae026374600add5b887", size = 4968631, upload-time = "2025-06-21T13:39:12.283Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/c3/b2/bc9c9196916376152d655522fdcebac55e66de6603a76a02bca1b6414f6c/pygments-2.20.0.tar.gz", hash = "sha256:6757cd03768053ff99f3039c1a36d6c0aa0b263438fcab17520b30a303a82b5f", size = 4955991, upload-time = "2026-03-29T13:29:33.898Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/c7/21/705964c7812476f378728bdf590ca4b771ec72385c533964653c68e86bdc/pygments-2.19.2-py3-none-any.whl", hash = "sha256:86540386c03d588bb81d44bc3928634ff26449851e99741617ecb9037ee5ec0b", size = 1225217, upload-time = "2025-06-21T13:39:07.939Z" },
+    { url = "https://files.pythonhosted.org/packages/f4/7e/a72dd26f3b0f4f2bf1dd8923c85f7ceb43172af56d63c7383eb62b332364/pygments-2.20.0-py3-none-any.whl", hash = "sha256:81a9e26dd42fd28a23a2d169d86d7ac03b46e2f8b59ed4698fb4785f946d0176", size = 1231151, upload-time = "2026-03-29T13:29:30.038Z" },
 ]
 
 [[package]]


mikelane (commented):
## Review Summary — PR #473

This PR clears all three known release blockers (#469, #468, #471) with a diff scoped exactly to what each issue asked for. Every claim in the PR description was independently verified rather than taken on faith.

### Verified independently

**`shell: bash` fix (#468) — confirmed live, not just by mechanism.** Pulled the actual job log for `Test (Python 3.11, windows-latest)` on this PR's CI run and the step header reads:

```
shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
```

That's the real `-e` fail-fast shell, not the old `pwsh -command` that swallowed pytest's exit code. All four `windows-latest` jobs on this PR are green — and for the first time in this repo's history that green is load-bearing rather than a byproduct of PowerShell discarding the pytest exit code. Traced through the script block: the only line with `||` (`coverage combine ... || echo "::warning..."`) is correctly guarded against `-e` aborting on an expected non-fatal condition; nothing else relies on continue-on-error semantics. Same pattern confirmed in `release.yml`. Both files' diffs are exactly the one-line `shell: bash` addition on the correct step, nothing else touched.

**`hasher.py` (#469) is genuinely docstring-only.** Diffed `395dfe8` against its parent: `hash_file` still reads `path.read_text(encoding='utf-8')` — the production fix stays with #465 as instructed. The re-scoped test (`it_matches_hash_file_output_to_hash_string_for_lf_source`) now writes explicit LF bytes via `write_bytes`, which is a real behavioral change from the old `write_text()` (platform-default newline) — not tautological, and it closes a latent Windows failure that would've appeared post-#465. The two `xfail(strict=False)` tests correctly pin the raw-byte-hash and CRLF≠LF-hash post-#465 contract; verified they currently report XFAIL (not XPASS) so the assertions are real, not vacuous.

**`pygments` bump (#471) is surgically clean.** `uv.lock` diff is exactly 3 insertions / 3 deletions — version, sdist hash, wheel hash for `pygments` only. `--ignore-vuln CVE-2026-4539` and its stale comment are gone. CI's Security Audit check passes.

**Regression-vs-main check (squash-merge hazard, mandatory for this phase):** `git merge-base HEAD origin/main` equals `origin/main`'s current tip — this branch is fully caught up with main, zero divergence, zero regression surface. Not applicable here beyond confirming there's nothing to check.

**Merge state:** `mergeStateStatus: CLEAN`, not draft, all 21 checks pass including Codecov.

### Minor observations (non-blocking)

1. **`hasher.py` docstring describes the post-#465 contract, not literally today's behavior.** As written today (pre-#465, `read_text()` with universal-newline translation), a CRLF file and its LF `hash_string` equivalent *do* still hash identically — the docstring's "not guaranteed for CRLF" claim only becomes true once #465 lands. This is clearly intentional (pre-documenting the narrower contract callers should rely on, ahead of an implementation change that's about to land), and it's disclosed in the commit message. Just flagging that a reader diffing this docstring against current code between now and #465 landing might be briefly confused. No action needed.

2. **`003fff1`'s commit message has an inverted ordering claim** ("this must land before #469") that the actual commit order contradicts (#469 landed first, correctly). The PR body explicitly calls this out and states the correct rationale. Whether that correction survives into the squash-merge commit message depends on this repo's "default to PR title and description for squash merges" setting — worth a glance before merging, but not worth blocking on; the correct context is preserved either in the PR body or the discussion either way.

### Praise

- The ordering discipline across all three commits (test-fix → CI-fix → dependency-fix) is deliberate and each commit message explains *why* that order matters, not just what changed.
- Explicitly not preempting #465's production fix in `hasher.py`, despite the temptation to "just fix it here," shows real discipline about PR scope.
- The `xfail(strict=False)` markers are a defensible choice given the constraint that neither `strict` setting can catch a bad #465 — the accompanying finding in the adversarial-qa artifact explains this precisely and this PR's tests match that explanation.

### Decision: **APPROVED**

0 CRITICAL, 0 HIGH, 0 MEDIUM blockers. Two LOW informational notes above, no action required before merge.

Issue #474, opened 2026-07-20
test: strict hermeticity enforcement reports a violation but does not fail the build


mikelane:
## Summary

Two defects, one nested inside the other:

1. **A `[SMALL]` test spawns a real subprocess**, violating hermeticity.
2. **`test_categories_enforcement = "strict"` reports that violation and does not fail the build.** The plugin literally prints `1 test failed due to violations` and pytest still exits `0`.

The second is the serious one. It means hermeticity enforcement has been decorative.

## Evidence

From run [29716150301](https://github.com/mikelane/pytest-gremlins/actions/runs/29716150301), **both** the `ubuntu-latest` and `windows-latest` 3.12 jobs:

```
======================== Hermeticity Violation Summary ========================
Violations detected (enforcement: strict):
  Network:     0 tests
  Filesystem:  0 tests
  Process:     1 test (tests/plugin/test_executor_cli.py::DescribeAutoExecutorResolution::it_resolves_auto_to_subprocess [SMALL])
  Database:    0 tests
  Sleep:       0 tests

Total: 1 violations in 1 test
1 test failed due to violations
```

And the same run's pytest summary:

```
==== 1522 passed, 2 skipped, 1 deselected, 2 xfailed in 198.79s ====
```

Zero failures. Job green.

This is **not** the Windows `pwsh` exit-code bug from #468 — it reproduces identically on Linux, which has always been able to fail. The step in this run executed under `bash --noprofile --norc -e -o pipefail`, so a non-zero pytest exit *would* have turned it red. pytest exited 0.

`pyproject.toml:268` confirms the setting:

```toml
test_categories_enforcement = "strict"
```

## Defect 1 — the violating test

`tests/plugin/test_executor_cli.py:97`:

```python
def it_resolves_auto_to_subprocess(self) -> None:
    """auto always resolves to subprocess until fork supports the full pipeline."""
    ...
    with (
        patch('pytest_gremlins.plugin._get_rootdir', return_value=Path('/project/src')),
        patch('pytest_gremlins.plugin._build_test_command', return_value=['pytest']),
        patch('pytest_gremlins.plugin._run_mutation_testing_inprocess') as mock_inprocess,
    ):
        _run_mutation_testing(mock_session, gremlin_session)
```

It patches the **in-process** path, but the whole point of the test is that `auto` resolves to the **subprocess** path — which is therefore left unpatched and actually spawns a process.

[...]

PR #483, opened 2026-07-20
test: finalize hasher test suite after #465 (closes #475)

### LINKED ISSUE #475: test: add salvaged hash_file edge-case tests (directory, empty, symlink, BOM)
## Added scope (2026-07-20): remove the now-redundant xfail markers — RELEASE-BLOCKING

PR #465 has merged, so the two `@pytest.mark.xfail` tests that #473 added to `tests/cache/test_hasher.py` now **always xpass** — the behavior they pin is live. An always-xpassing xfail gives zero regression protection while reading as coverage (the exact "passes for the wrong reason" trap that #469 existed to fix). Removing the markers converts them into real validators.

This half is **release-blocking** — the release gate will not clear until it's done. Folding it in here because it's the same file as the salvaged edge tests below.

**Do:**
- Remove the `@pytest.mark.xfail(...)` decorator (and the "remove the xfail marker once #465 merges" docstring lines) from both `it_hashes_file_content_as_raw_bytes` (line ~108) and `it_hashes_crlf_and_lf_files_differently_for_same_logical_content` (line ~127) in `tests/cache/test_hasher.py`.
- Confirm both now pass as normal tests (not xpass) against the merged `read_bytes` implementation.
- Also add the four salvaged edge-case tests described in the original issue body below.

After this lands, the only remaining release step is the CHANGELOG (drafted in project memory).

---

(original issue body follows)

## Summary

Four edge-case tests for `ContentHasher.hash_file` were written during the adversarial QA gate on PR #465 but never committed — they lived in an untracked file in a scratch worktree. None are covered by the tests added in #473. Recording them here so they aren't lost.

## Why they're worth having

`hash_file` derives cache keys. A defect there doesn't fail loudly — it silently poisons every incremental run (wrong cache hits → stale gremlin results → wrong mutation scores). It's a correctness chokepoint with thin edge-case coverage.

All four were verified to pass against **both** the current `read_text()` implementation and the `read_bytes()` version from #465, so they're portable regardless of merge order.

## The tests

```python
"""Edge-case tests for ContentHasher file hashing."""

import hashlib

import pytest

from pytest_gremlins.cache.hasher import ContentHasher


@pytest.mark.medium
class DescribeContentHasherFileErrors:
    """Tests for error behavior of hash_file."""

    def it_raises_is_a_directory_error_for_directories(self, tmp_path):
        """hash_file raises IsADirectoryError when given a directory."""
        hasher = ContentHasher()

        with pytest.raises(IsADirectoryError):
            hasher.hash_file(tmp_path)

    def it_hashes_empty_files(self, tmp_path):
        """hash_file returns the SHA-256 of empty bytes for an empty file."""
        hasher = ContentHasher()
        file_path = tmp_path / 'empty.py'
        file_path.write_bytes(b'')

        assert hasher.hash_file(file_path) == hashlib.sha256(b'').hexdigest()

    def it_follows_symlinks_to_target_content(self, tmp_path):
        """hash_file hashes the symlink target's content, not the link itself."""
        hasher = ContentHasher()
        target = tmp_path / 'target.py'
        target.write_bytes(b'def foo(): pass\n')
        link = tmp_path / 'link.py'
        link.symlink_to(target)

        assert hasher.hash_file(link) == hasher.hash_file(target)

    def it_hashes_utf8_bom_files_by_raw_bytes(self, tmp_path):
        """hash_file includes the UTF-8 BOM in the digest."""
        hasher = ContentHasher()
        content = b'\xef\xbb\xbfdef foo(): pass\n'
        file_path = tmp_path / 'bom.py'
        file_path.write_bytes(content)

        assert hasher.hash_file(file_path) == hashlib.sha256(content).hexdigest()
```

## Notes on each

- **`IsADirectoryError`** — `hash_file`'s docstring only promises `FileNotFoundError`. Passing a directory raises `IsADirectoryError` from `open()` under both implementations. Worth pinning, and arguably worth documenting in the docstring alongside `FileNotFoundError`.
- **Empty file** — `sha256(b'')` under both implementations (`read_text()` yields `''`, which encodes to `b''`).
- **Symlink** — both implementations follow the link and hash the target. Relevant because a repo with symlinked source would otherwise have undefined cache-key behavior.
- **UTF-8 BOM** — passes under both, but for different reasons: `read_bytes()` includes the BOM directly, while `read_text(encoding='utf-8')` decodes it to `﻿` and re-encodes to the same bytes. Worth a test precisely because that equivalence is non-obvious and would break under `utf-8-sig`.

## Not included

A fifth test, `it_matches_hash_string_for_crlf_files`, was deliberately discarded. It was a **reproduction** of the Windows failure that became #469, asserting a non-requirement (that CRLF bytes hash equal to `hash_string` of LF text). It can never legitimately go green — byte-exact hashing is the correct behavior — and #473 superseded it with tests that state the contract correctly.

## Acceptance criteria

- [ ] The four tests land in `tests/cache/` (fold into `test_hasher.py` or a sibling file)
- [ ] Suite green on all platforms
- [ ] Optionally: `hash_file` docstring mentions `IsADirectoryError`

## Provenance

Written by the adversarial-qa agent during the #465 gate. Low priority — pure coverage improvement, no known defect behind it.

mikelane:
Test-only follow-up now that PR #465 (byte-hashing) has merged. `tests/cache/test_hasher.py` only; `src/` untouched.

## Part 1 — remove two now-redundant xfail markers (was release-blocking)

#473 marked `it_hashes_file_content_as_raw_bytes` and `it_hashes_crlf_and_lf_files_differently_for_same_logical_content` as `@pytest.mark.xfail(strict=False)` because they pinned behavior from the then-unmerged #465. #465 has landed, so both were **always-xpassing** — an xfail that never fails gives zero regression protection while reading as coverage (the exact "passes for the wrong reason" trap #469 existed to fix).

Markers removed; both are now ordinary passing tests. The overall suite's `xpassed` count drops from 2 to **0**.

**These are genuinely load-bearing now** — the adversarial QA gate proved it by running each assertion against both implementations: under the merged `read_bytes()` both pass; under a `read_text().encode()` regression (pre-#465 behavior) both fail. So they would catch a regression, which is the whole justification for un-marking them.

## Part 2 — four salvaged edge-case tests

Added in a new `DescribeContentHasherFileErrors` class (matching the file's I/O-test convention). These were written during #465's original gauntlet but never committed:

- directory input → raises `OSError` (`IsADirectoryError` on POSIX, `PermissionError` on Windows — asserted at the portable `OSError` level after the Windows CI caught the platform difference)
- empty file → `sha256(b'')`
- symlink → hashes the target's content (real `symlink_to`)
- UTF-8 BOM → included in the digest (would fail under `utf-8-sig` stripping)

## Verification

- `pytest tests/cache/test_hasher.py -v` → 21 passed, **0 xfail/0 xpass**
- `pytest tests -m "small or medium and not wip"` → 1529 passed, 0 failed, 2 skipped, **0 xpassed** (was 2)
- `mypy src/pytest_gremlins` clean · `ruff check tests` + `format --check` clean

## Release

This was the last release-blocking item. After merge, the only remaining release step is the CHANGELOG entry.

Closes #475

### DIFF
--- a/tests/cache/test_hasher.py
+++ b/tests/cache/test_hasher.py
@@ -105,17 +105,8 @@ class DescribeContentHasherFileIO:
 
         assert file_hash == string_hash
 
-    @pytest.mark.xfail(
-        reason='passes once #465 lands: hash_file will hash raw bytes instead of read_text() output',
-        strict=False,
-    )
     def it_hashes_file_content_as_raw_bytes(self, tmp_path):
-        """hash_file equals the SHA-256 digest of the file's raw bytes.
-
-        Fails on current main because hash_file still hashes read_text()
-        output, which applies universal-newline translation. Remove the
-        xfail marker once #465 merges.
-        """
+        """hash_file equals the SHA-256 digest of the file's raw bytes."""
         hasher = ContentHasher()
         file_path = tmp_path / 'test.py'
         file_path.write_bytes(b'class MyClass:\r\n    pass\r\n')
@@ -124,17 +115,8 @@ class DescribeContentHasherFileIO:
 
         assert result == hashlib.sha256(file_path.read_bytes()).hexdigest()
 
-    @pytest.mark.xfail(
-        reason='passes once #465 lands: hash_file will hash raw bytes instead of read_text() output',
-        strict=False,
-    )
     def it_hashes_crlf_and_lf_files_differently_for_same_logical_content(self, tmp_path):
-        """A CRLF file and an LF file with the same logical content hash differently.
-
-        Fails on current main because read_text()'s universal-newline
-        translation makes CRLF and LF files hash identically. Remove the
-        xfail marker once #465 merges.
-        """
+        """A CRLF file and an LF file with the same logical content hash differently."""
         hasher = ContentHasher()
         logical_content = 'class MyClass:\n    pass\n'
         lf_path = tmp_path / 'lf.py'
@@ -175,6 +157,45 @@ class DescribeContentHasherFileIO:
         assert str(file1) in result
         assert str(file2) in result
         assert result[str(file1)] != result[str(file2)]
+
+
+@pytest.mark.medium
+class DescribeContentHasherFileErrors:
+    """File I/O edge cases for ContentHasher — require real filesystem access."""
+
+    def it_raises_is_a_directory_error_for_directories(self, tmp_path):
+        """hash_file raises IsADirectoryError when given a directory."""
+        hasher = ContentHasher()
+
+        with pytest.raises(IsADirectoryError):
+            hasher.hash_file(tmp_path)
+
+    def it_hashes_empty_files(self, tmp_path):
+        """hash_file returns the SHA-256 of empty bytes for an empty file."""
+        hasher = ContentHasher()
+        file_path = tmp_path / 'empty.py'
+        file_path.write_bytes(b'')
+
+        assert hasher.hash_file(file_path) == hashlib.sha256(b'').hexdigest()
+
+    def it_follows_symlinks_to_target_content(self, tmp_path):
+        """hash_file hashes the symlink target's content, not the link itself."""
+        hasher = ContentHasher()
+        target = tmp_path / 'target.py'
+        target.write_bytes(b'def foo(): pass\n')
+        link = tmp_path / 'link.py'
+        link.symlink_to(target)
+
+        assert hasher.hash_file(link) == hasher.hash_file(target)
+
+    def it_hashes_utf8_bom_files_by_raw_bytes(self, tmp_path):
+        """hash_file includes the UTF-8 BOM in the digest."""
+        hasher = ContentHasher()
+        content = b'\xef\xbb\xbfdef foo(): pass\n'
+        file_path = tmp_path / 'bom.py'
+        file_path.write_bytes(content)
+
+        assert hasher.hash_file(file_path) == hashlib.sha256(content).hexdigest()
 
 
 @pytest.mark.small
--- a/tests/cache/test_hasher.py
+++ b/tests/cache/test_hasher.py
@@ -163,11 +163,16 @@ class DescribeContentHasherFileIO:
 class DescribeContentHasherFileErrors:
     """File I/O edge cases for ContentHasher — require real filesystem access."""
 
-    def it_raises_is_a_directory_error_for_directories(self, tmp_path):
-        """hash_file raises IsADirectoryError when given a directory."""
+    def it_raises_os_error_for_directories(self, tmp_path):
+        """hash_file raises OSError when given a directory.
+
+        IsADirectoryError is raised on POSIX systems, PermissionError on
+        Windows. Both are subclasses of OSError, so we assert the portable
+        contract without specifying the platform-specific subclass.
+        """
         hasher = ContentHasher()
 
-        with pytest.raises(IsADirectoryError):
+        with pytest.raises(OSError):  # noqa: PT011
             hasher.hash_file(tmp_path)
 
     def it_hashes_empty_files(self, tmp_path):


mikelane (commented):
## Review verdict: APPROVE

**Scope check** — diff touches only `tests/cache/test_hasher.py` (+41/-20); `src/` is untouched. Confirmed via `gh pr view --json files` and `git diff origin/main...HEAD`.

### Part 1 — xfail removal
- Verified no `xfail`/`xpass` markers remain anywhere in the file (`rg xfail|xpass` → 0 hits).
- Docstrings on `it_hashes_file_content_as_raw_bytes` and `it_hashes_crlf_and_lf_files_differently_for_same_logical_content` no longer reference #465 as unmerged — the "Fails on current main because..." paragraphs were removed cleanly, no dangling references.
- **Independently re-verified the load-bearing claim** (didn't just trust the adversarial-qa artifact): simulated the pre-#465 `read_text().encode()` implementation against both tests' assertions.
  - Raw-bytes test: fails under the old impl (mismatch detected) ✅
  - CRLF/LF test: old impl hashes both files identically, so the `!=` assertion would fail ✅
  - Confirms these two tests genuinely catch a regression back to the old behavior; unmarking them was correct, not vacuous.

### Part 2 — new `DescribeContentHasherFileErrors` tests
- Directory → `IsADirectoryError`, empty file → `sha256(b'')`, symlink → target content, UTF-8 BOM → included in digest. All four assert on real behavior, non-tautological, `@pytest.mark.medium` matches the sibling `DescribeContentHasherFileIO` class's convention (real filesystem I/O).
- Naming/style matches file conventions: declarative `it_*` names, single quotes, no "should".

### Tooling
- `ruff check` + `ruff format --check` on the file: clean.
- `pytest tests/cache/test_hasher.py -v`: 21 passed, 0 xfail/xpass — matches PR description.

### Regression check (branch staleness)
- `git merge-base HEAD origin/main` == current `origin/main` HEAD (`be1cedc`) — the branch is not stale, it's built directly on current main. No squash-merge regression risk.

### Non-blocking observation (LOW)
- `it_follows_symlinks_to_target_content` is the first use of `Path.symlink_to()` in this test suite. Windows runners can restrict symlink creation without Developer Mode/elevated privileges — worth watching once the Windows CI legs finish (at review time, the one Windows failure visible was an unrelated "Install uv" infra flake, not a test failure; other Windows Python versions were still pending). Not a blocker — if it turns out flaky on Windows, a follow-up `sys.platform` skip/xfail would be a one-line fix.

### Not in scope
- Pre-existing hermeticity violation in `tests/plugin/test_executor_cli.py` (issue #474) — this PR doesn't touch that file.

**0 CRITICAL, 0 HIGH, 0 MEDIUM, 1 LOW (non-blocking).** Verdict artifact written to `~/second-brain/06_Agent/artifacts/pr-review/mikelane__pytest-gremlins__483.json` with `verdict: APPROVE`.

(Posted as COMMENT rather than a formal GitHub approval — PR author and reviewing account are the same GitHub user, and GitHub does not permit self-approval.)

mikelane (commented):
## Re-review at 9882977 (supersedes ae4ad08 verdict)

Note: posting as COMMENT, not APPROVE — GitHub blocks self-approval on owner-authored PRs. The verdict artifact (written separately, `verdict: "APPROVE"`, 0/0/0/0 findings) is what gates the merge hook.

The prior review approved this PR at `ae4ad08` with one LOW watch-item: whether the new symlink test would behave on Windows CI. Windows CI did fail — but on a different, unflagged test: `it_raises_is_a_directory_error_for_directories` asserted `pytest.raises(IsADirectoryError)`, which is POSIX-only (opening a directory on Windows raises `PermissionError`, not `IsADirectoryError`). The flagged symlink test actually passed on Windows.

### Delta verified (`ae4ad08..9882977`)

Exactly one commit, one file, `tests/cache/test_hasher.py`:
- `it_raises_is_a_directory_error_for_directories` → `it_raises_os_error_for_directories`
- `pytest.raises(IsADirectoryError)` → `pytest.raises(OSError)  # noqa: PT011`
- Docstring now documents the POSIX/Windows split explicitly

`src/` remains untouched across the whole PR (both commits) — confirmed via `gh pr diff --name-only`.

### Why the broadened assertion is still meaningful

`ContentHasher.hash_file` (`src/pytest_gremlins/cache/hasher.py:59`) is a bare `path.read_bytes()` — no directory guard in the implementation. The only way this raises anything for a directory input is the OS-level open failure, which is `IsADirectoryError` on POSIX and `PermissionError` on Windows — both `OSError` subclasses, no others possible for this specific failure mode. The test still genuinely locks in "passing a directory raises an error," just without over-specifying the platform-dependent subclass. Not vacuous.

`# noqa: PT011` is justified: `match=` would have to encode platform/locale-specific OS error text ("Is a directory" vs "Permission denied" vs any localized variant), which is precisely the coupling this fix is removing. Suppressing PT011 here with a comment explaining why is the correct call, not a shortcut.

### Empirical confirmation

`gh pr checks 483`: 21/21 pass, including all four `windows-latest` legs (3.11–3.14). `mergeStateStatus: CLEAN`.

### Everything from the prior verdict still holds

No xfail/xpass markers remain in the file, the other edge-case tests (empty files, symlink-follows-target, UTF-8 BOM) are unchanged, and the two previously-verified regression tests are intact.

**Decision: APPROVED** at `988297790b6e7de1abed3f97f7c31dfceb88e42d`. 0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW. No blockers.

mikelane (commented):
## Review: PR #483

**Verdict: APPROVE** (0 CRITICAL, 0 HIGH, 0 MEDIUM, 1 LOW)

Self-authored PR — posting as COMMENT since GitHub blocks self-approval; the pr-reviewer verdict artifact is the durable gate.

### Scope verification
- Diff vs `origin/main` touches exactly `tests/cache/test_hasher.py` (+46/-20). `src/` genuinely untouched. Confirmed via `git diff origin/main...HEAD --stat` and `gh api .../pulls/483/files`.
- Branch is fully current with `main` — `git merge-base HEAD origin/main` shows 0 commits on `main` since divergence. No squash-merge staleness risk (n/a here, but checked per house rule for any branch age).

### Part 1 — xfail removal is genuinely load-bearing
Read `src/pytest_gremlins/cache/hasher.py`: `hash_file` does `hashlib.sha256(path.read_bytes()).hexdigest()`. Confirmed the two newly-unmarked tests (`it_hashes_file_content_as_raw_bytes`, `it_hashes_crlf_and_lf_files_differently_for_same_logical_content`) would fail under a `read_text().encode()` regression — CRLF would get universal-newline-translated to LF before hashing, breaking both assertions. These are not decorative; removing `xfail(strict=False)` converts genuine dead coverage into a real regression guard, exactly as the PR body claims.

### Part 2 — four new edge-case tests
All four (`it_raises_os_error_for_directories`, `it_hashes_empty_files`, `it_follows_symlinks_to_target_content`, `it_hashes_utf8_bom_files_by_raw_bytes`) are non-tautological, use real filesystem operations (`tmp_path`, real `symlink_to`), and match the file's existing `DescribeContentHasherFileIO`-style conventions in a new `DescribeContentHasherFileErrors` class under `@pytest.mark.medium`.

The directory test's `pytest.raises(OSError)` (with `# noqa: PT011`) is not vacuous: `hash_file` opens the path via `read_bytes()`, so for an existing, user-owned directory the only reachable failure is the directory-open error — `IsADirectoryError` on POSIX, `PermissionError` on Windows, both `OSError` subclasses. `FileNotFoundError` and other `OSError` variants can't fire on a real, existing directory. The `noqa` is justified: narrowing via `match=` would couple the assertion to platform-specific error text ("Is a directory" vs "Permission denied").

**Empirical cross-platform proof**: `gh pr checks 483` shows all 21 checks passing, including all four `windows-latest` legs (Python 3.11–3.14). `mergeStateStatus` is `CLEAN`.

### Quality gates
- `pytest tests/cache/test_hasher.py -v` → 21 passed, 0 xfail/xpass locally, matching the PR's claim.
- `ruff check` / `ruff format --check` on the file → clean.
- `mypy src/pytest_gremlins` → clean.
- Adversarial-QA gauntlet artifact at this exact head SHA (`988297790b6e7de1abed3f97f7c31dfceb88e42d`) shows `status: pass`, 0 bugs, 0 blockers.

### LOW finding
- **`tests/cache/test_hasher.py:166-176`** — The PR description states the directory test asserts "`IsADirectoryError` (exact type, not a broad `OSError` catch)", but the actual code asserts the broader `OSError` (correctly, for Windows portability — `PermissionError` isn't a subclass of `IsADirectoryError`). The description appears to predate the cross-platform fix and wasn't updated. Not a code defect — the code is correct — just a stale PR body claim worth fixing before merge so the description matches what shipped.

### Praise
Good discipline turning a known always-xpassing gap into a real regression guard rather than just deleting the marker silently, and choosing the portable `OSError` assertion over a platform-specific one once CI surfaced the Windows failure mode — that's exactly the kind of fix that would otherwise resurface as a flaky Windows job later.

